PR #39, opened 2014-12-22
fixes for OpenNI 1.x depth&rgb reg/sync

nazcaspider:
Simultaneous Rgb and Depth was broken with OpenNI 1.x. Only depth channel was readable. Rgb node had to be configured after depth. Also FrameSync is now enabled.


stevenlovegrove:
I've cherry-picked the two useful commits. Thanks Renato.

PR #42, opened 2015-04-15
Bugfix: GrabNewest now calls GrabNewest instead of GrabNext

janjachnik:
I couldn't understand why my video stream was lagging when my main loop was running below frame-rate - this seems to be the reason.


stevenlovegrove:
Thanks Jan. I've cherry-picked this on devel (4941503378d3b86ee0b379e8f2f0d0d6bf3ca703)

PR #56, opened 2015-11-01
FFMPEG: fixed removed libav definitions.

### LINKED ISSUE #27: ffmpeg issue?
Recently, I have my Ubuntu changed to 14.04 LTS.
The FFmpeg is not available in this version of linux. so 
$ sudo apt-get install ffmpeg
does not work. 

I can find following.
http://fcorti.com/2014/04/22/ffmpeg-ubuntu-14-04-lts/

Have you ever try to run Pangolin 14.04?

magniffico:
Some of the old definitions in the code are no longer usable with recent version of ffmpeg (removed by [this commit](http://git.videolan.org/?p=ffmpeg.git;a=commit;h=c10caea2144fefc5e3e80807f301b3bf7a9b2753) on 2015/09/05). This commit should fix it.
Fixes #27

### DIFF
--- a/include/pangolin/video/drivers/ffmpeg.h
+++ b/include/pangolin/video/drivers/ffmpeg.h
@@ -91,7 +91,7 @@ protected:
     AVPacket        packet;
     int             numBytesOut;
     uint8_t         *buffer;
-    PixelFormat     fmtout;
+    AVPixelFormat     fmtout;
 };
 
 enum FfmpegMethod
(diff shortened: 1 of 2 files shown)


stevenlovegrove:
Thanks for this patch, @magniffico . This compiles fine for me too. Do you have any idea if this might break for anyone on an Ubuntu distro from the last few years?

magniffico:
Looks like the AV_ prefix was introduced with [this ffmpeg commit](http://git.videolan.org/?p=ffmpeg.git;a=commit;h=78071a1420b425dfb787ac739048f523007b8139) on 2012/10/06. Ubuntu 12.04 LTS is using libav in-place of ffmpeg and the [same commit on libav](https://git.libav.org/?p=libav.git;a=commit;f=libavutil/pixfmt.h;h=78071a1420b425dfb787ac739048f523007b8139) was made on the same day. This did not make it to 0.8 release that Ubuntu 12.04 uses. There should be no problem with Ubuntu versions from 14.04. Not sure about the no longer supported ubuntu releases.

PR #57, opened 2015-11-02
Include <vector> in common header file.

crheckman:
Including `<vector>` in a source file versus a common header was causing a build failure with `src/image/image_io.cpp` since commit d431d5f0b30106f249b76a51d84018c0ef3648a5.

### DIFF
--- a/include/pangolin/image/image_common.h
+++ b/include/pangolin/image/image_common.h
@@ -34,6 +34,7 @@
 #include <exception>
 #include <string>
 #include <map>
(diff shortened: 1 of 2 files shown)


stevenlovegrove:
I prefer to keep includes in the relevant source file to minimise compilation time and avoid namespace pollution etc, but I've hopefully fixed your issue with commit 45912a4.

crheckman:
Sounds good! Thanks!

PR #73, opened 2016-01-26
Add shared memory buffer video driver

visigoth:
These changes add a video driver that sources its data from an external process via a shared memory buffer. The external process writes to the frame buffer and signals a condition variable when the buffer is full. The video driver waits on the condition variable to synchronize the "frame rate" of the external application.

### DIFF
--- a/examples/SharedMemoryVideoWriter/main.cpp
+++ b/examples/SharedMemoryVideoWriter/main.cpp
@@ -1,0 +1,62 @@
+#include <pangolin/pangolin.h>
+
+#include <pangolin/utils/condition_variable.h>
+#include <pangolin/utils/shared_memory_buffer.h>
+#include <pangolin/utils/timer.h>
+
+#include <cmath>
+#include <cstdint>
+
+// This sample acts as a soft camera. It writes a pattern of GRAY8 pixels to the
+// shared memory space. It can be seen in Pangolin's SimpleVideo sample using
+// the shmem://example?size=640x480 video URI.
+
+using namespace pangolin;
+
+uint8_t generate_value(basetime t, basetime start)
+{
+  int64_t us = TimeDiff_us(start, t);
+
+  // 10s sinusoid
+  double d = 10./(M_PI) * ((double)us/1000000);
+  d = std::sin(d);
+  d = d*128 + 128;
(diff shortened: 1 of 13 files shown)


examples/SharedMemoryVideoWriter/main.cpp
```diff
@@ -0,0 +1,62 @@
...
+#include <pangolin/utils/condition_variable.h>
+#include <pangolin/utils/shared_memory_buffer.h>
+#include <pangolin/utils/timer.h>
+
+#include <cmath>
+#include <cstdint>
+
+// This sample acts as a soft camera. It writes a pattern of GRAY8 pixels to the
+// shared memory space. It can be seen in Pangolin's SimpleVideo sample using
+// the shmem://example?size=640x480 video URI.
```
stevenlovegrove: Pangolin Video URI's are of the form scheme:[option1=value1,option2=value2,...]//device_name

include/pangolin/utils/condition_variable.h
```diff
@@ -0,0 +1,27 @@
+#pragma once
+
+#include <pangolin/utils/timer.h>
+
+#include <memory>
```
stevenlovegrove: Pangolin uses Boost if C++11 isn't available. We can't use unique_ptr, but shared_ptr is available via a compatibility header (https://github.com/stevenlovegrove/Pangolin/blob/master/include/pangolin/compat/memory.h)

src/utils/semaphore.cpp
```diff
@@ -0,0 +1,83 @@
+#include <pangolin/utils/semaphore.h>
+
+#include <fcntl.h>
+#include <semaphore.h>
+#include <sys/stat.h>
```
stevenlovegrove: Looks like this isn't going to build on Windows. Not sure about OSX. We will need to conditionally add this stuff depending on platform.

stevenlovegrove:
This looks good Shaheen. We just need to avoid C++11 and add some platform config. I don't mind doing that myself so that you can get on and start using this.

stevenlovegrove:
I just tested it on OSX and it works great!

visigoth:
created a new development branch for myself, so closing this one and opening a new PR

PR #78, opened 2016-01-29
Shared memory video driver

visigoth:
This PR creates a virtual camera driver that can be driven via POSIX shared memory. A producer creates a shared memory buffer and a condition variable to signal that it is full. A consumer can create the `SharedMemoryVideo` driver and poll via `VideoInterface`

### DIFF
--- a/examples/SharedMemoryVideoWriter/main.cpp
+++ b/examples/SharedMemoryVideoWriter/main.cpp
@@ -1,0 +1,64 @@
+#include <pangolin/pangolin.h>
+
+#include <pangolin/compat/memory.h>
+#include <pangolin/utils/posix/condition_variable.h>
+#include <pangolin/utils/posix/shared_memory_buffer.h>
+#include <pangolin/utils/timer.h>
+
+#include <cmath>
+
+// This sample acts as a soft camera. It writes a pattern of GRAY8 pixels to the
+// shared memory space. It can be seen in Pangolin's SimpleVideo sample using
+// the shmem:[size=640x480]//example video URI.
+
+using namespace pangolin;
+
+unsigned char generate_value(basetime t, basetime start)
+{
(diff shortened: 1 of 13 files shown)


src/video/drivers/shared_memory.cpp
```diff
@@ -0,0 +1,64 @@
...
+SharedMemoryVideo::~SharedMemoryVideo()
+{
+}
+
+void SharedMemoryVideo::Start()
+{
+}
+
+void SharedMemoryVideo::Stop()
+{
```
visigoth: would it be important to stop servicing `GrabNext` if `Stop` is called?

examples/SharedMemoryVideoWriter/main.cpp
```diff
@@ -0,0 +1,62 @@
...
+  d = std::sin(d);
+  d = d*128 + 128;
+  return static_cast<uint8_t>(d);
+}
+
+int main(int argc, char *argv[])
+{
+  std::string shmem_name = "/example";
+
+  auto shmem_buffer = create_named_shared_memory_buffer(shmem_name, 640 * 480);
```
stevenlovegrove: auto is C++11

src/utils/posix/condition_variable.cpp
```diff
@@ -0,0 +1,88 @@
...
+  pthread_cond_init(&pthread_data->cond, &cattr);
+
+  ptr.reset(static_cast<ConditionVariableInterface *>(
+      new PThreadConditionVariable(move(shmem))));
+  return ptr;
+}
+
+boostd::shared_ptr<ConditionVariableInterface>
+open_named_condition_variable(const string &name) {
+  auto shmem = open_named_shared_memory_buffer(name, true);
```
stevenlovegrove: auto

Issue #80, opened 2016-02-05
runtime error when using orb slam v2


jhz-linhe:
I am trying orb slam v2, encountering below error when launching.
Parallels ubuntu 14.04 virtual machine on Mac, 8G Memory, 4 Cores. 

terminate called after throwing an instance of 'std::runtime_error'
  what():  Pangolin X11: Unable to retrieve framebuffer options
Aborted (core dumped)

Thank you all in advance for any help!

stevenlovegrove:
This would appear to be a duplicate of issue #74

Issue #74, opened 2016-01-27
Pangolin X11: Unable to retrieve framebuffer options


sunstarchan:
Hello, 

when I follow these steps to run ORB_SLAM2(https://github.com/raulmur/ORB_SLAM2), I get following error. 

terminate called after throwing an instance of 'std::runtime_error'
  what():  Pangolin X11: Unable to retrieve framebuffer options
Aborted (core dumped)

I'm running it with Ubuntu running in Parallels on Mac. 
What's the main reason for this issue?

Thanks

sunstarchan:
@stevenlovegrove one more comment, I tried to build the project with the default C++ compiler on Mac, it  built successfully, and all examples could run too. but when I tried to build with gcc/g++, the Object C code for Mac has build errors, such as 

                 from /Users/Ming/Documents/Projects/package_dir/Pangolin/src/display/device/display_osx.mm:33:
/usr/include/objc/NSObject.h:22:4: error: 'instancetype' does not name a type
 - (instancetype)self;
    ^
/usr/include/objc/NSObject.h:36:4: error: 'instancetype' does not name a type
 - (instancetype)retain OBJC_ARC_UNAVAILABLE;
    ^
/usr/include/objc/NSObject.h:38:4: error: 'instancetype' does not name a type
 - (instancetype)autorelease OBJC_ARC_UNAVAILABLE;
    ^
/usr/include/objc/NSObject.h:60:4: error: 'instancetype' does not name a type
 - (instancetype)init
    ^
/usr/include/objc/NSObject.h:66:4: error: 'instancetype' does not name a type
 + (instancetype)new OBJC_SWIFT_UNAVAILABLE("use object initializers instead");
    ^
/usr/include/objc/NSObject.h:67:4: error: 'instancetype' does not name a type
 + (instancetype)allocWithZone:(struct _NSZone *)zone OBJC_SWIFT_UNAVAILABLE("use object initializers instead");
    ^
/usr/include/objc/NSObject.h:68:4: error: 'instancetype' does not name a type
 + (instancetype)alloc OBJC_SWIFT_UNAVAILABLE("use object initializers instead");
    ^
In file included from /System/Library/Frameworks/Foundation.framework/Headers/NSArray.h:5:0,
                 from /System/Library/Frameworks/Foundation.framework/Headers/Foundation.h:10,
                 from /System/Library/Frameworks/Cocoa.framework/Headers/Cocoa.h:12,
[...]

PR #81, opened 2016-02-05
Do not set GLX_SAMPLES when asking for framebuffer options

paulinus:
Using on virtual machines people are getting the following error:

```
terminate called after throwing an instance of 'std::runtime_error'
what(): Pangolin X11: Unable to retrieve framebuffer options
```

see (https://github.com/stevenlovegrove/Pangolin/issues/74, https://github.com/stevenlovegrove/Pangolin/issues/80, https://github.com/raulmur/ORB_SLAM2/issues/4 and https://github.com/raulmur/ORB_SLAM2/issues/15)

This PR solves the problem by not setting the GLX_SAMPLES option when asking for framebuffer options as done in this tutorial as it is done in this [tutorial](https://www.opengl.org/wiki/Tutorial:_OpenGL_3.0_Context_Creation_(GLX)).  As far as i understand, the GLX_SAMPLE_BUFFERS and GLX_SAMPLES are set later on the same function anyway

### DIFF
--- a/src/display/device/display_x11.cpp
+++ b/src/display/device/display_x11.cpp
@@ -120,8 +120,8 @@ int CreateX11Window(
         GLX_DEPTH_SIZE      , 24,
         GLX_STENCIL_SIZE    , 8,
         GLX_DOUBLEBUFFER    , glx_doublebuffer ? True : False,
-        GLX_SAMPLE_BUFFERS  , glx_sample_buffers,
-        GLX_SAMPLES         , glx_sample_buffers > 0 ? glx_samples : 0,
+        // GLX_SAMPLE_BUFFERS  , glx_sample_buffers,
+        // GLX_SAMPLES         , glx_sample_buffers > 0 ? glx_samples : 0,
         None
     };
 


mp3guy:
Does using it like this not have the same effect? 

https://github.com/mp3guy/ElasticFusion/blob/master/GUI/src/Tools/GUI.h#L44

stevenlovegrove:
The issue with commenting those lines is that they specify the minimum requirements for enumeration. This means that you will end up with the worst framebuffer available. We could instead return the best, but this results in performance issues. You're running into this problem because your display does not support the minimum framebuffer requirements specified by the application (as per Tom's comment). I've been thinking about a fix, and the best option is probably for Pangolin to enumerate all options (as per your PR), but then to select the best option less than or equal to the requested samples. This will then be more forgiving for software renderers etc.

stevenlovegrove:
I've just pushed 45b38c15fae075aec0e3e865780b5235b4055848 which will hopefully address this issue whilst keeping the current behaviour on systems with decent graphics support. Please comment further if that still does not work for anyone.

paulinus:
That works great. Thanks @stevenlovegrove and sorry for the dirty PR ;-)

One detail is that i needed to add librt to the link libraries to build the devel branch (https://github.com/stevenlovegrove/Pangolin/compare/devel...paulinus:add-librt?expand=1)

stevenlovegrove:
Thanks for the heads up. I've applied your patch in the conditional that uses librt: e4a79aa2a0c43a89da5bb27b2f40b19b5415d73c. Hope that solves it.

paulinus:
perfect, thanks!

sakurazhu:
i have use the method you suggested as "e4a79aa"  but  it did not work on my computer .
i use ubuntu 14.04  on VMware® Workstation10.0.4 .  i do not know how to fix it . Could you give me any help ?  @stevenlovegrove

stevenlovegrove:
@sakurazhu To clarify, you see this behaviour in the devel branch? I don't have any concrete suggestions, but you should try playing with the visual_attribs variable here https://github.com/stevenlovegrove/Pangolin/blob/devel/src/display/device/display_x11.cpp#L97 to see if you can find a mode supported by your renderer. I'm not overly familiar with X11, so you may have to do some googling. You could also see if VMware has any options for enabling OpenGL acceleration or something.

sakurazhu:
Tthank you for your suggestion. I have solved the problem  by changing my dataset's path and the path of asssociation. 

2016-05-28 

myworldpig 



发件人：Steven Lovegrove <notifications@github.com>
发送时间：2016-05-28 00:22
主题：Re: [stevenlovegrove/Pangolin] Do not set GLX_SAMPLES when asking for framebuffer options (#81)
收件人："stevenlovegrove/Pangolin"<Pangolin@noreply.github.com>
抄送："zhuying"<myworldpig@163.com>,"Mention"<mention@noreply.github.com>

@sakurazhu To clarify, you see this behaviour in the devel branch? I don't have any concrete suggestions, but you should try playing with the visual_attribs variable here https://github.com/stevenlovegrove/Pangolin/blob/devel/src/display/device/display_x11.cpp#L97 to see if you can find a mode supported by your renderer. I'm not overly familiar with X11, so you may have to do some googling. You could also see if VMware has any options for enabling OpenGL acceleration or something.

AlexandrGraschenkov:
I use Parallels Desktop on Mac with installed Ubuntu 14.04. Problem solved after disable 3d acceleration on virtual machine settings. After this works laggy but works.

Kaju-Bubanja:
Can you describe how to turn off 3d acceleration? And is this only relevant when running in a virtual machine or could I try to turn of hardware acceleration like [here](https://unix.stackexchange.com/questions/408582/how-to-disable-hardware-acceleration-in-linux)?

PR #85, opened 2016-02-20
seems to fix #expect directive problem

mp3guy:


stevenlovegrove:
Applied in this commit, aafe689a92192f44ca0eb73aa94eadf70850af8b. Thanks.

PR #88, opened 2016-03-04
Explicitly link against libpthread

visigoth:
The POSIX dependencies introduced for shared memory buffers and condition variables placed an implicit dependency on libpthread, which would usually be resolved by other library imports. This makes the dependency explicit.


stevenlovegrove:
The patch contained an erroneous -l and I'd rather keep the dependency at the point of the code that requires it, so hopefully fixed in bcfa21e.

PR #93, opened 2016-03-27
Add pleora get_temperature flag.

renzodenardi:
Adding flag to enable/disable sensor temperature polling since it is a blocking call that can take several ms.


renzodenardi:
Bundled up with pleora threaded.

PR #106, opened 2016-04-26
New thread filter

renzodenardi:
Filter that allows running any of the child as a separate thread.

### DIFF
--- a/include/pangolin/video/drivers/thread.h
+++ b/include/pangolin/video/drivers/thread.h
@@ -1,0 +1,99 @@
+/* This file is part of the Pangolin Project.
+ * http://github.com/stevenlovegrove/Pangolin
+ *
+ * Copyright (c) 2014 Steven Lovegrove
+ *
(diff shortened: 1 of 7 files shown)


include/pangolin/video/drivers/thread.h
```diff
@@ -0,0 +1,234 @@
...
+                BufPType bp = validBuffers.front();
+                validBuffers.pop_front();
+                return bp;
+            }
+        }
+    }
+
+    void addValidBuffer(BufPType bp) {
+        // Add buffer to valid buffers queue.
+        std::lock_guard<std::mutex> vlock(vMtx);
```
stevenlovegrove: std::lock_guard is C++11. Is this in boost too? Can you use the boostd style workaround?

renzodenardi: lock_guard and unique_lock are also part of boost so I simply added a compat/locks.h header to take care of the c++11 or boost inclusion following other examples in boostd.

PR #110, opened 2016-05-07
Command line support for repeated options.

renzodenardi:
Pleora added support for generic command line commands through syntax Command=Execute
Params.h allows repeated commands but Get will always return the last value for the key (command line options are read in left to right order).

### DIFF
--- a/src/video/drivers/pleora.cpp
+++ b/src/video/drivers/pleora.cpp
@@ -26,7 +26,7 @@
  */
 
 #include <pangolin/video/drivers/pleora.h>
-#include <unistd.h>
+#include <pangolin/compat/thread.h>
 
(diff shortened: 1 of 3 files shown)


src/video/drivers/pleora.cpp
```diff
@@ -26,6 +26,7 @@
  */
 
 #include <pangolin/video/drivers/pleora.h>
+#include <unistd.h>
```
stevenlovegrove: unistd.h isn't available on Windows. What is this inclusing for?

renzodenardi: is for usleep, let me use the cross platform pangolin timer...
good catch.

renzodenardi:
Review comment addressed.

PR #114, opened 2016-05-12
Make pipe video input/output much more robust

visigoth:
These commits make pipe input and output much more robust to applications disappearing on either end. With these changes, I was able to have sample applications start/stop and have them pick right back up where they left off.

The first couple of commits add some infrastructure to be used later. `PacketStreamReader` is refactored a bit next. Then, reader and writer code paths are both made more robust.

`PipeOpen` leaks file descriptors, but closing the file descriptor within `PipeOpen` would cause a reader to believe the pipe was empty and that the writer had disappeared. Instead, the file descriptor is kept open until higher level code can figure out what to do with it. Because `PangoVideoOutput` now opens the file before closing this extra file descriptor, the reader doesn't get signaled. Also, it was possible to get stuck in a loop in `FlushPipe` because `read` can return `0` bytes read.

On the reader side, several changes were made within `PacketStreamReader` to be more robust. Format errors still result in exceptions, but if data could not be read, it now fails more gracefully. In `PangoVideo`, a graceful failure or exception from a pipe is treated as a closed pipe. `PacketStreamReader` is closed and `GrabNext` becomes a proxy for checking whether the pipe has data in it. The pipe is opened in two stages: first create a file descriptor, then check whether the pipe has data in it and hydrate `PacketStreamReader`. It is done this way so that the writer knows there is a reader waiting for input. If, instead, the file descriptor was opened, tested, and immediately closed, a "smart" writer that checks whether the pipe is opened for reading would likely never start writing to the pipe again.

### DIFF
--- a/include/pangolin/utils/file_utils.h
+++ b/include/pangolin/utils/file_utils.h
@@ -74,9 +74,6 @@ PANGOLIN_EXPORT
 bool IsPipe(const std::string& file);
 
 PANGOLIN_EXPORT
-bool PipeOpen(const std::string& file);
-
-PANGOLIN_EXPORT
 bool IsPipe(int fd);
 
 PANGOLIN_EXPORT
--- a/include/pangolin/utils/file_utils.h
+++ b/include/pangolin/utils/file_utils.h
@@ -77,7 +77,7 @@ PANGOLIN_EXPORT
 bool IsPipe(int fd);
 
 PANGOLIN_EXPORT
-bool PipeOpenForRead(const std::string& file);
+int WritablePipeFileDescriptor(const std::string& file);
 
 PANGOLIN_EXPORT
 int ReadablePipeFileDescriptor(const std::string& file);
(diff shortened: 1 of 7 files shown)


include/pangolin/utils/file_utils.h
```diff
@@ -125,6 +126,78 @@ inline std::string ToLowerCopy( const std::string& str )
...
+            if (_filebuf->is_open()) {
+                _filebuf->close();
+            }
+            delete _filebuf;
+            _filebuf = nullptr;
+        }
+    }
+
+private:
+    __gnu_cxx::stdio_filebuf<std::ifstream::char_type> *_filebuf;
```
stevenlovegrove: It doesn't look hopeful that this will build on Windows. We need to guard anything platform dependent.

stevenlovegrove:
This looks great - You missed a few windows guards but I've added those and pushed to devel

PR #116, opened 2016-05-13
Check if we are in playback before trying a seek.

renzodenardi:


stevenlovegrove:
Resolved in 06ca099, thanks Renzo.

PR #122, opened 2016-06-13
Added the ability to remove variables again, see SimpleDisplay example.

martinruenz:
In many scenes it might be necessary to display temporary information and in this case it would be helpful to add Variables to a panel and to later delete them again.

This feature has been added to Pangolin.


martinruenz:
Since I am only superficially familiar with the Pangolin source, please double-check that I didn't introduce a memory leak somewhere.

martinruenz:
I only now saw the devel-branch - sorry. If applicable I will create a new request on the correct branch.

PR #123, opened 2016-06-13
Added the ability to remove variables (Var<..>) again

martinruenz:
In many scenes it might be necessary to display temporary information and in this case it would be helpful to add Variables to a panel and to later delete them again.

This feature has been added to Pangolin. The behaviour is illustrated in the SimpleDisplay example: pressing the button now deletes and adds new Variables.

Since I am only superficially familiar with the Pangolin source, please double-check that I didn't introduce a memory leak somewhere.

### DIFF
--- a/examples/SimpleDisplay/main.cpp
+++ b/examples/SimpleDisplay/main.cpp
@@ -33,13 +33,13 @@ void SampleMethod()
 
 
 int main( int argc, char* argv[] )
-{  
+{
   // Load configuration data
   pangolin::ParseVarsFile("app.cfg");
 
   // Create OpenGL window in single line
   pangolin::CreateWindowAndBind("Main",640,480);
-  
+
   // 3D Mouse handler requires depth testing to be enabled
   glEnable(GL_DEPTH_TEST);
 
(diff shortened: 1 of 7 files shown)


stevenlovegrove:
Thanks for this. I'll take a look when I get a chance.

NikolausDemmel:
I just gave this a shot and it rebases (almost) cleanly on current master.

A quick test shows that it works with normal variables and attached variables (adding and removing and adding again).

Looking at the code, I'm not sure if memory management is correct. What happens of there are multiple var-objects referencing the same variable and then it is removed? Are all VarValue objects  destroyed properly after a variable is removed both for the attached and non-attached version?

NikolausDemmel:
PS: This was my test:

```
pangolin::Var<std::function<void(void)>> test_btn{
    "ui.test", []() {
      static bool flag = true;
      static int bar = 20;

      pangolin::Var<bool> foovar{"ui.foo", true, true};
      if (flag) {
        pangolin::Var<int>::Attach("ui.bar", bar, 0, 100);
      } else {
        foovar.destroy();
        pangolin::Var<int> barvar("ui.bar");
        barvar.destroy();
      }
      flag = !flag;
    }};
```

NikolausDemmel:
Just wanted to mention that I have been using this for a while and it has been working great to add and remove GUI elements in panels.

> Looking at the code, I'm not sure if memory management is correct. What happens of there are multiple var-objects referencing the same variable and then it is removed? Are all VarValue objects destroyed properly after a variable is removed both for the attached and non-attached version?

But I guess the above concern is not yet investigated and the way I use it, I'm not sure I cover all possible cases.

stevenlovegrove:
It's been a few years now, but I've just pushed this feature to the master branch (84569fa5ec75b74b9b79b24012dda2a70ca4ac26). I've waited mainly until my Var shared_ptr refactor was complete and I was confident about the lifetime semantics. I didn't base my implementation on this PR because I wanted to take the opportunity to revise the Var callback mechanism as well. Hopefully this fulfills your needs!

PR #124, opened 2016-06-14
Load modelview matrix directly

visigoth:
Load the OpenGL model view matrix directly rather than multiplying it on the stack


stevenlovegrove:
Cherry picked in 024fe51e90c21f6aa9420063bdca969eda72c976

PR #125, opened 2016-06-23
Shared contexts

visigoth:
This set of commits makes the following changes:

- Adds a copy constructor for `GlBufferCudaPtr`
- Improves tracking of the current thread's active PangolinGl context
- Adds explicit context sharing which is necessary on Windows because after a context has been used, it is no longer shareable. It's possible to have the creation code path silently fail when sharing is not permitted, but I think this is better.


visigoth:
Removed ShareContexts and added a getter as a hack for code outside of pangolin for now.

PR #138, opened 2016-08-23
Export RegisterScheme so that clients can use the function

visigoth:
This method needs to be exported so that clients can use it.


stevenlovegrove:
I've rebased these changes in, thanks.

PR #143, opened 2016-08-26
Realsense driver

jstraub:
Support for realsense RGBD sensors via librealsense (added FindLibRealsense.cmake as well to find the library). Tested on my Ubuntu 14.04 with current librealsense https://github.com/IntelRealSense/librealsense.


stevenlovegrove:
This is great, thanks!

PR #152, opened 2016-10-06
RenderVboIbo: fix number of vertex indices for glDrawElements

christian-rauch:
This commit fixes an issue in `RenderVboIbo` where the incorrect number of indices (`ibo.num_elements`) is used for `glDrawElements`. The correct number of indices to use is `ibo.num_elements*ibo.count_per_element`.


christian-rauch:
I am using this fix for quite some time to correctly render meshes.
Can I have some feedback if this is the desired solution and what is needed to get the fix merged upstream?

stevenlovegrove:
Sorry, I haven't had a chance to look properly at it. The GL_TRIANGLE_STRIP parameter should probably be an option, not hard-coded. GL_TRIANGLE_STRIP is more efficient generally than GL_TRIANGLES if you have managed to set up your indices in a way to use them. Regarding the element count, it seems like a usage issue - I need to take a look at it carefully because your change would appear to break the semantics for existing code.

stevenlovegrove:
If you look at the documentation for glDrawElements (https://www.opengl.org/sdk/docs/man4/html/glDrawElements.xhtml), the second parameter is the "number of elements to be rendered". If you take MakeTriangleStripIboForVbo as an example, you would typically have only 1 'count per element' for an IBO. What would multiple counts mean, in an index buffer?

christian-rauch:
The problem is probably my mesh, it has holes. E.g. `MakeTriangleStripIboForVbo` cannot generate a triangle strip with missing triangles.

Here is example code (with corrupted mesh): [triangle_strip_example.tar.gz](https://github.com/stevenlovegrove/Pangolin/files/637476/triangle_strip_example.tar.gz) that compares both approaches (with and without `MakeTriangleStripIboForVbo` preprocessing).

Running:
`./triangle_strip_example nostrip ../bottle_holes.ply` will render the mesh (GL_TRIANGLES) with holes as it is e.g. rendered by MeshLab: 
mesh_holes


`./triangle_strip_example strip ../bottle_holes.ply` will generate the triangle-strip and render (GL_TRIANGLE_STRIP) wrongly:
mesh_strip

There should be an optional flag to render meshes that have not been converted to a triangle strip. Alternative this could be determined by the `count_per_element`, which will be 1 for a triangle strip and e.g. 3 for triangular faces.

PR #154, opened 2016-10-07
Enabled Wextra flag and fixed several warnings

RealRecon:
These commits enable the Wextra flag and fix most warnings related to it.
Despite a warning related to the Py_INCREF macro in src/python/PyModulePangolin.cpp tthat cannot be fixed, Pangolin compiles without any noise. I may missed some warnings since I have not installed all but the most important libraries Pangolin makes use of.


stevenlovegrove:
Thanks Patrick

PR #155, opened 2016-10-08
Enable C++11 and remove boost dependency

RealRecon:
Pangolin makes more and more use of C++11 features (e.g. nullptr, range-based for-loops, etc.). Therefore building with CPP11_NO_BOOST = OFF will fail compiling with tons of errors.

These commits force C++11 support by removing all boost related stuff. Since there is the convenient boostd namespace that is either replaced by std or boost depending on the state of CPP11_NO_BOOST, porting away from boostd is done here by replacing boostd with std and using the correct headers. Some headers in compat are now obsolete and removed as well.

I have taken special care to not forget any occurrences of boostd and the obsolete headers, but I may have missed some places. Please double check if everything is fine and no errors were introduced with this pull request.


stevenlovegrove:
I'm surprised to hear that Pangolin doesn't build without C++11. The drone.io build server is meant to be checking that (https://drone.io/github.com/stevenlovegrove/Pangolin). Any use of range-based-for loops and nullptr etc should be trivially fixable. Is there a good reason to be removing the non-C++11 configuration?

RealRecon:
I am using Pangolin as a locally included library in my CUDA project on Ubuntu 14.04 together with some other ones. Even version 0.3 had those problems but at a smaller degree. At that time I fixed this myself. But as I recently tried to upgrade to 0.4, more errors were thrown and fixing them became more difficult. With an upgrade to CUDA 8.0 (I skipped 7.5 where C++11 support was already introduced), everything is working fine now without any workarounds from my side.

I think it is about time to switch to C++11. CUDA supports this since one year, GCC 6 defaults to C++14 and many projects and libaries have made the step already. My project also greatly benefit since I was able to drop boost. Having less dependencies is always a goal one should aim for.

I think since nobody has complained so far about compiliation problems when disabling C++11 support, most likely he/she has not changed the default configuration meaning that he/she uses C++11. So I think there is no reason not to go for it.

RealRecon:
Regarding drone.io: It only checks for -DCPP11_NO_BOOST=1, thus with C++11 enabled. This explains why drone.io has not complained for more than one year (problems existed in 0.3 already).

But this highlights that it is about time for C++11. Nobody complained so far, so I expect that this step would not break any user code since the non-C++11 setups are obviously broken.

stevenlovegrove:
It actually does multiple builds. It runs this:

uname -a
gcc -v
sudo apt-get update || true

mkdir build
cd build

sudo apt-get install libboost-dev libboost-thread-dev libboost-filesystem-dev libglu-dev libglew-dev ffmpeg libavcodec-dev libavutil-dev libavformat-dev libswscale-dev libdc1394-22-dev libraw1394-dev

cmake -DCPP11_NO_BOOST=1 ..
make
rm -rf *

cmake -DBUILD_PANGOLIN_GUI=false ..
make
rm -rf *

cmake -DBUILD_PANGOLIN_VIDEO=false ..
make
rm -rf *

cmake -DBUILD_PANGOLIN_VARS=false ..
make
rm -rf *

cmake -DBUILD_PANGOLIN_VARS=false -DBUILD_PANGOLIN_VIDEO=false ..
make
rm -rf *

RealRecon:
Yes, of course. But since CPP_NO_BOOST is set to ON per default in CMakeLists.txt and the only call of cmake with this option set uses -DCPP11_NO_BOOST=1, all builds actually run with enabled C++11.

stevenlovegrove:
Okay, I gave in :)

RealRecon:
Thank you, Steven :)

PR #159, opened 2016-10-11
Fix typo in include directives

gergondet:
Hi,

This PR fixes a typo that was introduced when removing boost dependencies


stevenlovegrove:
Thanks!

PR #164, opened 2016-10-22
Fixed missing cmath include

starfys:
```include/pangolin/gl/colour.h``` makes references to ```floor``` and ```sqrt``` but does not include ```cmath`` for these references to work.


stevenlovegrove:
Thanks!

PR #166, opened 2016-10-29
Suppressed unused-parameter warnings

RealRecon:
When building with all optional dependencies disabled, some unused parameter warnings occur. This commit fixes that by suppressing them using the defined `PANGOLIN_UNUSED` macro that works for all compilers.


stevenlovegrove:
Thanks.

PR #167, opened 2016-10-30
Fixed some Appveyor warnings

RealRecon:
When looking into the logs and messages by Appveyor, there are a couple of warnings emitted. While most of them are related to external libraries like libjpg or caused by an overly picky Visual C++, there are two warnings that should be fixed.

The first commit is related to a warning about an implicit cast of a pointer to `bool`. Although it is not really important, fixing it by a direct check against a null pointer also clarifies the intention of `IsOpen` a bit.

The other commit is more critical since the user might pass a negative row number to an image object and would get a pointer to an illegal memory position. Thanks to the warning about an implicit cast of `size_t` to `int`, this is now fixed.

### DIFF
--- a/include/pangolin/image/image.h
+++ b/include/pangolin/image/image.h
@@ -78,12 +78,12 @@ struct Image {
         return Image<To>(w,h,pitch, (To*)ptr);
     }
 
-    T* RowPtr(int r)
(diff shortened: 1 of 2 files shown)


stevenlovegrove:
Thanks!

PR #169, opened 2016-11-03
Travis warnings

RealRecon:
Because Pangolin now requires C++11, I observed that we can drop the `PANGOLIN_OVERRIDE` macro and use `override` directly as is done by some files already. The other two commits fix some warnings I encountered in the logs of Travis related to the OSX builds.


stevenlovegrove:
Thanks!

PR #171, opened 2016-11-07
OpenNI2 cropping

tschmidt23:
added cropping to OpenNI2 driver and added option for turning on fast crop mode.

with this change, a crop of an OpenNI2 stream is taken by adding the symbol '#' to the stream string followed by an ImageRoi of the form X+Y+WxH. the fast cropping mode can be activated with the option 'fastcrop'. for example:

openni2:[img1=depth:640x480@30,img2=rgb:1280x1024@30#320+240+640x480,fastcrop=true,framesync=true]//


stevenlovegrove:
Thanks Tanner.

PR #174, opened 2016-11-14
adding optional range and length to markers

strasdat:
simple_plot_with_box_markers


stevenlovegrove:
Reimplemented in ee52494db3ead48e9dc1ddc135da635a0b98da9c. Thanks.

Issue #176, opened 2016-11-19
Error install


haythemsgh:
Hi,
I want to install Pangolin because it is required to install ORB SLAM2.
I installed all the Required Dependencies for Pangolin but after I hit cmake -j I get this error:

/home/haythem/Pangolin/src/video/drivers/merge.cpp: In constructor ‘pangolin::MergeVideo::MergeVideo(std::unique_ptr<pangolin::VideoInterface>&, const std::vector<pangolin::Point>&, int, int)’:
/home/haythem/Pangolin/src/video/drivers/merge.cpp:43:54: error: ‘assert’ was not declared in this scope
     assert(stream_pos.size() == src->Streams().size());
                                                      ^
make[2]: *** [src/CMakeFiles/pangolin.dir/video/drivers/merge.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [src/CMakeFiles/pangolin.dir/all] Error 2
make: *** [all] Error 2

haythemsgh:
I succeeded to solve this by just adding #include <assert.h> in merge.cpp

drinking-tea:
@haythemsgh hi haythemsgh, I have the exactly problem with you, can I ask which file did you #include in merge.cpp?

haythemsgh:
@drinking-tea Hi drinking-tea , you need to add #include<assert.h> in merge.cpp

PR #177, opened 2016-11-24
include assert.h in video/drivers/merge.cpp

TenninYan:
fixed issues #176

I had this problem using ubuntu 14.04 but not on ubuntu 16.04.


stevenlovegrove:
Ah, I just pushed a commit with this fix myself. Thanks.

PR #178, opened 2016-11-25
Fix bug: stack around the variable 'T_wc' was corrupted

maxint:
The bug is reproduced in win64. The variable 'T_wc' is composed of 12 double's, but `T_wc[12]` is wrote in function `LieSE3from4x4<>()`. Please check the code directly.

B. R.
maxint


stevenlovegrove:
Oops! You're right - thanks!

maxint:
You're welcome. Thanks for such handy OpenGL library!

PR #182, opened 2016-12-01
Two tiny modification

maxint:
1. Generate binary files only when any of binary files is newer than output file.
2. Add required headers for MSVC14 when HAVE_EIGEN is undefined.


stevenlovegrove:
I've rebased these changes on master. Thanks!

maxint:
You're welcome. Rebasing is great for linear commit flow 😄

PR #190, opened 2016-12-13
Fix librealsense include path finding

zhang-xin:
In code it's #include <librealsense/rs.hpp> but in CMake the path with subdir librealsense is saved. So if the upper dir is not added by others (/usr/local/include or any non-default path that librealsense was installed in), the compile will fail.

### DIFF
--- a/CMakeModules/FindLibRealSense.cmake
+++ b/CMakeModules/FindLibRealSense.cmake
@@ -16,9 +16,8 @@ if(${CMAKE_VERSION} VERSION_LESS 2.8.2)
 endif()
 
 #add a hint so that it can find it without the pkg-config
-find_path(LIBREALSENSE_INCLUDE_DIR rs.h
-    HINTS /usr/include/  /usr/local/include
-    PATH_SUFFIXES librealsense)
+find_path(LIBREALSENSE_INCLUDE_DIR librealsense/rs.h
+    HINTS /usr/include/  /usr/local/include)
 #add a hint so that it can find it without the pkg-config
 find_library(LIBREALSENSE_LIBRARY
     NAMES librealsense.so


stevenlovegrove:
Does it work if instead of your chance, you modify
https://github.com/zhang-xin/Pangolin/blob/ce7ded2a0701609401b9f9df308d12540e6891a7/CMakeModules/FindLibRealSense.cmake#L19

to

find_path(LIBREALSENSE_INCLUDE_DIR librealsense/rs.h

and remove the PATH_SUFFIXES bit?

zhang-xin:
Yes it works too and is clearer. Made the change as you suggested. Thanks.

alfred316:
Can someone explain to me how to fix this on the cmake-gui? I'm not clear on this answer, I'm sorry. 
But for context, I'm getting the same issue where it's saying it can't find librealsense/rs.hpp but I have the path to that in cmake in the realsense_include_dir which is /usr/local/include/librealsense2 and realsense_library which is /usr/local/include/librealsense2/rs.hpp

what am I missing here?

Thanks in advance!

zhang-xin:
@alfred316 it's not cmake issue, I guess you're using librealsense2 (librealsense2/rs.hpp), while this project is using librealsense 1.x (called librealsense legacy on project README: https://github.com/IntelRealSense/librealsense/tree/v1.12.1).

If you're using new released RealSense D400 series you may need to fix the code yourself, otherwise just use librealsense legacy.

alfred316:
@zhang-xin thank you for the link. I actually just found this a little bit ago and I'm trying to use it, but for some reason I'm running into the same issue. What's the cmake set up look like? since it has c links and realsense paths to set up, I'm wondering if I'm doing it wrong. For the realsense_include_dir I have it looking at the include/librealsense folder and for the library its looking at include/librealsense/rs.hpp and it still says it cant find librealsense/rs.hpp

I used snap to install it and its at /snap/librealsense/3/include/librealsense/rs.hpp

what am I missing still?

alfred316:
@zhang-xin i made a mistake its actually at /snap/librealsense/3/usr/include/librealsense/rs.hpp

alfred316:
@zhang-xin also one last thing, how do I run "make install" in a build folder in librealsense and keep it on version 1? it keeps changing to version 2 when I run make install, which I assume is just grabbing the latest version.

zhang-xin:
@alfred316 I don't think "make install" will change librealsense version, if you checkout v1.12.1 the code should always be 1.12.1. I didn't use snap so cannot help you on that.

The easiest way to make it work and not to modify cmake setting:
- clone librealsense project
- checkout to tag v1.12.1
- mkdir build; cd build; cmake ..; make; sudo make install

Then librealsense v1 is installed in /usr/local, when you build Pangolin cmake will find everything.

alfred316:
@zhang-xin I tried doing exactly that but the sudo make install command for some reason updated my librealsense to librealsense2

I used the files that snap installed and it recognizes the .hpp file it seems but now I have a new issue. Its giving me all kinds of errors apparently regarding the compiler. I tried to cmake -std=c++11 .. to compile it with c++11 like it seemed like it wanted but its still giving me a wall of errors all stemming from the compiler issue apparently. any idea what could be causing this or what I'm doing wrong? 

Thank you for taking the time to answer me and help me.

zhang-xin:
@alfred316 You're welcome:) That's weird, maybe you can check the file timestamp to see if it's what you installed.

To pass compiler c++11 option in cmake, add this to a proper place in your CMakeLists.txt file:

add_compile_options(--std=c++11)

alfred316:
@zhang-xin I tried putting add_compile_options(--std=c++11) in the cmakelists.txt file in the root directory but I'm still getting:
/usr/include/c++/5/bits/c++0x_warning.h:32:2: error: #error This file requires compiler and library support for the ISO C++ 2011 standard. This support must be enabled with the -std=c++11 or -std=gnu++11 compiler options.
 #error This file requires compiler and library support \
Which is followed by a plethora of errors for things in the rs.hpp file all saying they're not in scope or things aren't defined etc. Is this because I downloaded the librealsense from snap? 

I put the compiler thing at the top of the file before the first set of functions after the cmake minimum and project setting. Is that the wrong place? Am I missing something else?

zhang-xin:
@alfred316 first I should type -std=c++11 not --std=c++11, though on my machine g++ allows --std too.

Just checked Pangolin has already added compiler option for c++11 https://github.com/stevenlovegrove/Pangolin/blob/30c971e46aa22dfa44fbe50439dfa62cc151b63e/CMakeLists.txt#L50

so actually you don't need to add it yourself.

Maybe it's your OS/tool setup issue? FYI I'm using Ubuntu 16.04 x64 which g++ version is 5.4.0, just cloned Pangolin and built without any error.

alfred316:
@zhang-xin I'm also using ubuntu 16.04 x64 and have g++ 5.4.0.
I cloned and built Pangolin and it did without error. 

Would this be something in my link paths in cmake or something of the sort that I set up incorrectly? I'm just getting errors on almost every line of rs.hpp telling me things like the variables are declared void or they're not in the std:: namespace and things like that. I'm searching everywhere but I can't find a solution to this problem right now.

zhang-xin:
@alfred316 you can build Pangolin successfully? i'm confused i thought you're having issues on building Pangolin since we are in Pangolin's PR thread. Then when do you see the errors?

I don't think it's link paths issue in cmake since the errors are about source code, if it's linking issue errors like 'unsolved symbol: ...' will occur. Since I don't know what you're building I cannot help here. Anyway you can 1) check the lib path found by cmake (/usr/local/lib/librealsense.so?) and include path (/usr/local/include) for librealsense; 2) clean cmake build dir to start fresh.

PR #191, opened 2016-12-22
Teli: Allow decimation

strasdat:
Previously, DecimationHorizontal==2, DecimationVertical==2 did not work correctly. This change will resize the width, height accordingly.


strasdat:
Closing, this is already fixed differently on HEAD...

PR #192, opened 2017-01-06
Fix Appveyor warnings

RealRecon:
I've found these warnings in the Appveyor log. The `size_t -> int` warning might be critical if the user tries to pass negative numbers. The `double -> float` warning is less problematic, but should be fixed too.


RealRecon:
Updated the PR to also fix `size_t -> unsigned int` warnings from the newest commits.

PR #234, opened 2017-04-13
Fixed std::pair<T,T> MinMax() function

theobslhc:
Removed const keyword preventing std::pair from being updated and throwing read-only assignment error


stevenlovegrove:
Thanks!

PR #237, opened 2017-04-20
+ sudo make install

kozuch:
I guess this is needed for proper install? Without it the cmake script of https://github.com/JakobEngel/dso did not recognize Pangolin as being installed properly...

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,7 @@ mkdir build
 cd build
 cmake ..
 make -j
+sudo make install
 ```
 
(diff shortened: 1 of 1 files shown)


stevenlovegrove:
Installing it is optional. Pangolin actually exports itself to the cmake package cache, so that is my preferred method of using it. The code in the examples will work externally too, even without installation:

e.g.
https://github.com/stevenlovegrove/Pangolin/blob/master/examples/HelloPangolin/CMakeLists.txt

PR #241, opened 2017-05-12
Add support for saving PPM and PGM images

RealRecon:
Pangolin is currently limited to only load PPM and PGM images but not to save them. This PR adds the missing support.


stevenlovegrove:
Thanks!

PR #261, opened 2017-07-08
Update README.md

agharsh11:
updated readme for problem of crashes with make -j


stevenlovegrove:
Thanks for this - I instead replaced the section with an alternate cross-platform build command

`cmake --build .`

which should not have the problem that you encountered.

PR #262, opened 2017-07-09
glDrawAxis with OpenGlMatrix

christian-rauch:
These commits enable `glDrawAxis` to set the reference frame from `pangolin::OpenGlMatrix`.


stevenlovegrove:
Great, thanks!

Issue #45, opened 2015-06-10
VideoException for improper OpenNI VideoMode when reading from oni file


dcanelhas:
When setting up an openni device for live capture using an openni2: uri string this is not a problem, but i find that using the provided interface in pangolin/video/drivers/openni2.h for reading from files i.e.
```
   // Open openni file
    OpenNiVideo2(const std::string& filename);
```
isn't working as it should

The following code: 
```
pangolin::OpenNiVideo2 * video;
std::string file = "/home/danielcanelhas/Video/RGBD/Captured.oni";
video = new pangolin::OpenNiVideo2(file);  
```
causes the following

terminate called after throwing an instance of 'pangolin::VideoException'
  what():  Couldn't set OpenNI VideoMode
		Stream setProperty(3) failed

This can be "fixed" by removing the check in openni2.cpp : 331

        if(rc != openni::STATUS_OK)
            throw VideoException("Couldn't set OpenNI VideoMode", openni::OpenNI::getExtendedError());

(this time I'm extra confident that I am actually on the devel branch)

I feel like I am doing it wrong though I couldn't find a parser for file: uri strings that handles oni extensions anywhere.  That would probably be the most logical way of handling it.

qazmichaelgw:
Hello dcanelhas, Can you send me the example to set up an openni device for live capture using an openni2? I am going to use the XBOX 360 to capture the data!
Best regards

stevenlovegrove:
You should have been able to use "openni2:///home/danielcanelhas/Video/RGBD/Captured.oni" as your URI string (as used by VideoViewer for example). Opening oni files should work. If this is still an issue for anyone, please re-open this issue and perhaps capture a small oni sequence for me to test on.

Algomorph:
It's still an issue.

```
./VideoViewer "openni2:///media/algomorph/Data/4dmseg/TestCapture.oni"
Couldn't set OpenNI VideoMode
		Stream setProperty(3) failed
```

Algomorph:
Test .oni file: [link](http://algomorph.com/storage/TestCapture.oni)

Algomorph:
The problem is that for some reason you cannot set the video mode on an OpenNI stream from an .oni file. The simple work-around is to replace line 330 in ```src/video/drivers/openni2.cpp``` with
```std::cout << "Warning: Couldn't set OpenNI VideoMode" << openni::OpenNI::getExtendedError();```

Minimum example to reproduce:
```C++
[...]

PR #263, opened 2017-07-12
Solution for #45

Algomorph:
For OpenNI2 driver, prevents setVideoMode when an .oni file uri is given, since that results in an OpenNI2 error. Solution tested for .oni files & on OpenNI device.


stevenlovegrove:
Thanks!

PR #281, opened 2017-08-11
Mouse Modifiers without GLUT (on windows)

randi120:
Hey @stevenlovegrove, 
I was developing on Ubuntu 16.04 and expanded a couple Handlers to use modifiers on mouse actions.

I was surprised to find no modifiers when I switched back to Windows 7, so I added a few that are GLUT independent. 

They follow the pattern of `glutGetModifiers`, namely CTRL, SHIFT, ALT (as bitmasks).


stevenlovegrove:
Ah, I guess I didn't get around to that :), Thanks for the patch! I'll try to get this merged in after I give it a test on Windows.

randi120:
It would be good to double check it on a system with GLUT. I think it should work :)

If it helps here a screen grab of SimpleDisplay with debug print statements on `Handle3D` SimpleDisplay

stevenlovegrove:
Rebased! Thanks!

randi120:
:)

PR #302, opened 2017-11-03
include definition of Imf::IStream

christian-rauch:
Fixes the missing class definition:
```
Pangolin/src/image/image_io_exr.cpp:35:31: error: invalid use of incomplete type ‘class Imf::IStream’
 class StdIStream: public Imf::IStream
```


stevenlovegrove:
Thanks!

PR #304, opened 2017-11-17
representing matrix as Eigen::Affine transformation

christian-rauch:
OpenGlMatrix constructor and operator to interface with Eigen's affine transformation.


stevenlovegrove:
cherry-picked 4bd344914aecce7ba497c1e9f56a5960f8a544ed Thanks!

PR #305, opened 2017-11-20
Add package.xml for catkin_tools.

NikolausDemmel:
Not sure if you are willing to include this?

With this, you can build Pangolin as part of a catkin workspace
with catkin_tools. Note that with this, Pangolin is still a pure
cmake package, not a catkin package. For a catkinized pangolin,
see https://github.com/uzh-rpg/pangolin_catkin.

You can test it for example with

```
mkdir -p /tmp/test-ws/src
cd /tmp/test-ws/src
git clone https://github.com/NikolausDemmel/Pangolin
cd /tmp/test-ws
catkin init
catkin build

# now pangolin is installed in /tmp/test-ws/devel and catkin packages in the ROS ws can find_package it.
```

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0"?>
+<package format="2">
+  <name>pangolin</name>
+  <version>0.5.0</version>
+  <description>pangolin</description>
+  <maintainer email="stevenlovegrove@gmail.com">Steven Lovegrove</maintainer>
+  <author     email="stevenlovegrove@gmail.com">Steven Lovegrove</author>
+  <license>MIT</license>
+  <buildtool_depend>cmake</buildtool_depend>
+  <depend>libglew-dev</depend>
+  <depend>cmake</depend>
+  <depend>python</depend>
+  <export>
+    <build_type>cmake</build_type>
+  </export>
+</package>


christian-rauch:
I also would like to see this. For now, I am maintaining my own patches to have Pangolin built by catkin.

Maybe we could already include `<build_type>ament_cmake</build_type>` for the ament build system that is going to be used for ROS2?

NikolausDemmel:
Interesting, can you add both build_type exports at the same time and have it work with catkin_tools and ament?

christian-rauch:
Sorry, my previous suggestion was wrong. I just checked and you can only provide a single `<build_type>`. catkin and ament will complain if more than one is defined.

You can only build plain cmake packages with catkin and ament if `<build_type>cmake</build_type>` is defined.

NikolausDemmel:
Ah I see. I can confirm that dropping Pangolin with the package.xml from my branch in an ament workspace works as expected.

stevenlovegrove:
I'm happy to add this if it's useful, but I don't know how to maintain it. Is it still worth me putting this in mainline?

NikolausDemmel:
> I'm happy to add this if it's useful, but I don't know how to maintain it. Is it still worth me putting this in mainline?

Thanks for your consideration! I would say yes, it is worth adding. The information in there should be quite stable. E.g. if new dependencies are not added straight away to the `package.xml`, its still usable (just install dependencies manually, it won't affect the build). If something is actually broken, I think it is ok if you rely on the people that use it to provide a patch.

So I would be happy if you add it and never think about it again until the next PR that concerns it.

stevenlovegrove:
cherry-picked 42d8751fa2dd1e4a5213acebc67f80d5760f6ead Thanks!

PR #306, opened 2017-11-21
fixed OSX default oversized font problem

dshayden:
OSX default font rendering has been a problem for multiple OSX versions, causing the UI panel font to spill over and be unreadable on all the example projects, (on a Thunderbolt display, resolution 2560 x 1440 for instance). This effectively only changes the font height assignment default from 30 to 15, causing fonts to look fine.

### DIFF
--- a/src/display/device/display_osx.mm
+++ b/src/display/device/display_osx.mm
@@ -57,7 +57,7 @@ namespace pangolin
 
 WindowInterface& CreateWindowAndBind(std::string window_title, int w, int h, const Params& params )
 {
-    const bool is_highres = params.Get<bool>(PARAM_HIGHRES, true);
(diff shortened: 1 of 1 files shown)


stevenlovegrove:
This isn't quite the right fix. I use Pangolin on my macbook pro for example, and I want 'high-res' set to true. The differentiator is whether osx thinks the screen is a 'retina' display or not. I don't know if you know how to query that?

stevenlovegrove:
You'll have be happy to know that this is now configurable by an environment variable (38c4edf1a9dd3f748fcdb85e8bfa0ff374e88a44) and the default is probably the setting that you want anyway. 

For anyone wanting the retina resolution without scaling, you can use
```
setenv PANGOLIN_WINDOW_URI "cocoa:[HIGHRES=true]//"
```

PR #311, opened 2017-12-08
bug fix for build error.  fatal error:librealsense/rs.hpp: No such fi…

cxzhp:
…le for directory


stevenlovegrove:
This addresses the same issue as #190 - I don't like polluting the include path so I'd rather put the parent directory on the path.

Are you able to check if my proposed change (in the linked PR) works instead?

Thanks for highlighting this issue!

stevenlovegrove:
Fixed in this commit: https://github.com/stevenlovegrove/Pangolin/commit/30c971e46aa22dfa44fbe50439dfa62cc151b63e

Thanks!

PR #315, opened 2017-12-22
fix a crash bug for GUI on OSX

guohengkai:
The `CreateWindowAndBind` function will cause crash on Mac OSX 10.12.6. This is cased by the member call `run_pre` and `run_step` for subclass `PangolinNSApplication` with base class pointer. As suggested in the [Documents](https://developer.apple.com/documentation/appkit/nsapplication), creating a custom `NSApplication` subclass is disfavor. So I modify them into static member functions.


stevenlovegrove:
Cherry-picked b6eb178bcbefda1ada1370872fa69fc7bf620b92 Thanks!

PR #328, opened 2018-02-04
add window factory

mnaza:


src/display/device/display_osx.mm
```diff
@@ -222,4 +223,23 @@ inline void FixOsxFocus()
...
+PANGOLIN_REGISTER_FACTORY(OsxWindow)
+{
+  struct OsxWindowFactory : public FactoryInterface<WindowInterface> {
+    std::unique_ptr<WindowInterface> Open(const Uri& uri) override {
+
+      const std::string window_title = uri.Get<std::string>("window_title", "window");
+      const int w = uri.Get<int>("w", 640);
+      const int h = uri.Get<int>("h", 480);
+      const bool is_highres = uri.Get<bool>("is_highres", true);
+      return std::unique_ptr<WindowInterface>(CreateOsxWindowAndBind(window_title, w, h, is_highres));
```
stevenlovegrove: I know I asked for this to return a unique_ptr, but this isn't really unique here (or in the other device implementations) as it has already been added to the global contexts map. I think it would be better to only add to the contexts map in the common CreateWindowAndBind code path (converting the unique_ptr to shared_ptr), with the device depended factory method just returning a genuine unique ptr. That would also allow you to avoid returning a raw pointer in CreateOsxWindowAndBind

mnaza: Correct this

src/ios/PangolinUIView.mm
```diff
@@ -183,5 +184,19 @@ void FinishFrame()
         throw std::runtime_error("pangolin::FinishFrame() Not supported on this platform");
     }
     
+    PANGOLIN_REGISTER_FACTORY(IosWindow)
+    {
+      struct IosWindowFactory : public FactoryInterface<WindowInterface> {
+      std::unique_ptr<WindowInterface> Open(const Uri& uri) override {
+        
+        const std::string window_title = uri.Get<std::string>("window_title", "window");
+        CreateAndroidWindowAndBind(window_title);
```
stevenlovegrove: Android?

mnaza: correct it

stevenlovegrove:
This has now made it into master. Thanks!

PR #339, opened 2018-03-12
Emscripten support

mnaza:


stevenlovegrove:
This looks great. I'm having some trouble building. Are you able to add an emscripten readme?

mnaza:
with installed `emscripten` to build just type next command in console:
```
mkdir build
cd build
emcmake cmake -DBUILD_TESTS=OFF ..
```

stevenlovegrove:
And how to launch in a webpage?

stevenlovegrove:
(also, I had to add BUILD_SHARED_LIBS=off to successfully compile)

mnaza:
to launch You may use simple html file with context:
```
<!doctype html>
    <head>
        <meta charset="utf-8">
        <title>test</title>
    </head>
    <body>
        <canvas id="canvas"></canvas>
        <script type="text/javascript">
            var canvas = document.getElementById("canvas");
            canvas.width = window.innerWidth;
            canvas.height = window.innerHeight;
            var Module = {};
            Module.canvas = canvas;
        </script>
        <script type="text/javascript" src="testfile.js"></script>
    </body>
</html>
``` 
where `testfile.js` is the file builded with emscripten

mnaza:
agree with `BUILD_SHARED_LIBS=off`, but my version of `emscripten` is ignore "sharing" in any case

mnaza:
and You can use `-DWASM=ON` for `cmake` to enable webAssembly build

stevenlovegrove:
I tried using your file, but I have a Cross origin request problem when launching from my machine. Do you have a recommendation for how I should test this locally?

`SimpleDisplay.js:1 Failed to load file:///home/stevenlovegrove/code/Pangolin/build/examples/SimpleDisplay/SimpleDisplay.js.mem: Cross origin requests are only supported for protocol schemes: http, data, chrome, chrome-...`

mnaza:
You need to use some kind of http server. in simple way try:
```
python3.4 -m http.server
```
and type in browser `127.0.0.1:8080`

stevenlovegrove:
Great tips, thanks! I now am able to run them.

I do see a problem in HelloPangolin and SimpleDisplay where the mouse input is not working correctly when moving the cube. In HelloPangolin, it jumps down and right when I click. In SimpleDisplay, it instantly dissapears when I drag. This happens in both chrome and firefox. Does it work correctly for you?

mnaza:
I don`t try in firefox, but in chromium works fine

mnaza:
I have try in Firefox, and have same problem. Now it is work correctly.

stevenlovegrove:
Which version of chrome? Does it behave for you exactly like the desktop application?

mnaza:
Default chromium from Debian repo

stevenlovegrove:
I see - you are unable to read the framebuffer? Can you think of a workaround for that? For navigation to work, it's quite important that we can read the depth value under the mouse.

stevenlovegrove:
Just looking and found https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/readPixels - that should let us have a full implementation, right?

mnaza:
Can You point me what part of code did You mean?

src/handler/handler.cpp
```diff
@@ -254,17 +254,22 @@ void Handler3D::MouseMotion(View& display, int x, int y, int button_state)
         }else if( button_state == MouseButtonLeft )
         {
             // Left Drag: in plane translate
-            if( ValidWinDepth(last_z) )
+#if !defined(__EMSCRIPTEN__)
```
stevenlovegrove: These guards aren't needed so long as last_z is setup correctly

src/handler/handler.cpp
```diff
@@ -134,7 +134,11 @@ void Handler3D::PixelUnproject( View& view, GLprecision winx, GLprecision winy,
 {
     const GLint viewport[4] = {view.v.l,view.v.b,view.v.w,view.v.h};
     const pangolin::OpenGlMatrix proj = cam_state->GetProjectionMatrix();
+#if defined(__EMSCRIPTEN__)
+    glUnProject(winx, winy, winz, (const float*)Identity4d, proj.m, viewport, &Pc[0], &Pc[1], &Pc[2]);
```
stevenlovegrove: This cast is incorrect. Identity4f should be used instead - I think this is why navigation was so crazy before.

src/plot/plotter.cpp
```diff
@@ -390,7 +390,7 @@ void Plotter::Render()
     if(extern_draw_function && show) {
         glMatrixMode(GL_PROJECTION);
         glLoadIdentity();
-        glOrtho(rview.x.min, rview.x.max, rview.y.max, rview.y.min, -1, 1);
+//        glOrtho(rview.x.min, rview.x.max, rview.y.max, rview.y.min, -1, 1);
```
stevenlovegrove: this changes the meaning of the draw call

stevenlovegrove:
I pushed a small fix on my version of your branch 279f41826dcf25c2b43a7be89b21beba75d2158b regarding the glUnProject issue.

Regarding glReadPixels, on line https://github.com/stevenlovegrove/Pangolin/blob/279f41826dcf25c2b43a7be89b21beba75d2158b/src/handler/handler.cpp#L154 the glReadPixels is not linking correctly (when we take it out of the GLES conditional). Since you have emscripten configured with GLES3, I'm not sure why the symbol isn't getting resolved.

mnaza:
correct this

stevenlovegrove:
Although it now links with this change, it doesn't function correctly and you can see in a javascript console some runtime errors. To begin with, I get a glReadBuffer function not found. I comment that out (since that just selects front or back buffer), and then I get:

`WebGL: readPixels: Unexpected format or type`

It doesn't seem to like GL_DEPTH_COMPONENT with GL_FLOAT. Using GL_RGBA and GL_UNSIGNED_BYTE works to read the color - I'm not sure how to get it to read the depth buffer. Any ideas?

In firefox, from the console, you can play with reading the RGBA color as follows:
`test = new ArrayBuffer(100);`
`testview = new Uint8Array(test);`
`GLctx.readPixels(320,240,1,1,0x1908,0x1401,testview)`

mnaza:
As I understand there is only undirect way to read depth buffer. Need tp render scene to FBO with ahder like `gl_FragColor = vec4(vec3(gl_FragCoord.z), 1.0);` and than `readPixels` with `GL_RGB`

Issue #236, opened 2017-04-20
Wayland support with GL ES2


gapaluec:
I am trying to add Wayland support for Pangolin. I started with Wayland & OpenGL and planning to move to Wayland & GL ES2 to get Pangolin working on small embedded devices which only support GL ES2.

Issue 1:
As of now I am facing issue with mouse integration. I am getting same response with middle click and left click. in SimpleDisplay example, I have tested that Handler3D::MouseMotion is getting called with correct mouse button mapping. Still response is same.

Issue 2: When compiling with GL ES2, only a blank window is appearing when I try to run SimpleDisplay.

Any pointers will be helpful. Codes are at https://github.com/gapaluec/Pangolin

stevenlovegrove:
This is great! As you can probably tell, I've not used the gles2 engine stuff for a while, so it may have gotten quite broken without testing.

I'm afraid I'm completely unfamiliar with Wayland, or even how to use it. Is there an easy way to get it working alongside X11 on Ubuntu for testing?

gapaluec:
With Ubuntu moving to Wayland ( https://wiki.ubuntu.com/Wayland ), we have more reason to add Wayland support to Pangolin.

NikolausDemmel:
@christian-rauch @pagaluec, you mention Qt  here. Is it possible to combine the Pangolin viewer with additional (independent) QtWidgets / Windows?

NikolausDemmel:
Ah ok, thanks.

Actually, what I would like to do is not having Qt widgets inside pangolin, but the other way round, i.e. using a pangolin window together with some other independent qt windows in the same application. I don't really know much about at which level the "window management", "event loops", or other aspects of both libraries might be conflicting. I guess I should just try it out.

stevenlovegrove:
It should theoretically work fine - this is why Pangolin has a BindToContext method, so that you can use a context that you have already been given (by e.g. a Qt OpenGL widget). You will need to hook up the Qt widget input and resize signals to call the process::Mouse(...) etc events in display.h

PR #348, opened 2018-04-16
Wayland + EGL support

christian-rauch:
I continued the work of @gapaluec and implemented the window interface for Wayland (https://github.com/stevenlovegrove/Pangolin/issues/236). I am now at the state where I can process all keyboard and mouse input, show the rendered surface and resize the window (including fullscreen via tab).
At the moment, the used window interface (X11 or Wayland) is selected at runtime. If a Wayland server is available (i.e. when running a Wayland compositor), the Wayland interface is selected. If no compositor is available, it falls back to X11. For this reason, the X11 and Wayland windows use the same `scheme` which I called `linux` for now.
This also introduces new dependencies on wayland, xkbcommon and EGL (`sudo apt install libwayland-dev libxkbcommon-dev libegl1-mesa-dev`).

There are some issues left:
1.  [`glewInit()`](https://github.com/stevenlovegrove/Pangolin/blob/c2a6ef524401945b493f14f8b5b8aa76cc7d71a9/src/display/display.cpp#L165) fails when using a glew version lower than 2, since older versions call GLX related functions (e.g. for Ubuntu 16.04). In this case, you need to activate a workaround by setting `FALLBACK_X11_IF_GLEW1`. I haven't seen other calls to glew functions in Pangolin. Is glew actually used in Pangolin? E.g. I can remove the glewInit() line and successfully use the Wayland interface with glew<2.
2. When running under Weston, `eglSwapBuffers` blocks at the second call. I.e. the second time `eglSwapBuffers` gets called, it does not return. This behaviour does not occur under Gnome or KDE.
3. A major issue is that there are no window decorations at the moment. I.e. there is no title bar and no window borders where you can grab and drag/resize windows. At the moment, neither Gnome or KDE support server side decorations in an official way. An immediate solution would be to draw the window decorations manually (client-side decorations). In the long-term, I hope that server-side decorations become available.

The first two issues are minor, since they do not occur on modern systems with standard desktops (tested on Ubuntu 18.04). But I could use some recommendations regarding issue 3. What do you think about using custom client-side decorations for Wayland in Pangolin?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ dist: trusty
 
 before_install:
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sudo apt-get -qq update ; fi
-  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sudo apt-get install libeigen3-dev libglew-dev libc++-dev; fi
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sudo apt-get install libeigen3-dev libglew-dev libc++-dev libwayland-dev libxkbcommon-dev libegl1-mesa-dev; fi
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then pyenv versions && pyenv global system 3.6; fi
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update          ; fi
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install eigen glew ; fi
--- a/include/pangolin/gl/colour.h
+++ b/include/pangolin/gl/colour.h
@@ -66,13 +66,13 @@ struct Colour
     }
 
     /// Construct from component values
-    inline Colour(float red, float green, float blue, float alpha = 1.0f)
+    inline Colour(const float red, const float green, const float blue, const float alpha = 1.0f)
         : red(red), green(green), blue(blue), alpha(alpha)
     {
     }
 
     /// Construct from rgba array.
-    inline Colour(float rgba[4])
+    inline Colour(const float rgba[4])
     {
         r = rgba[0];
         g = rgba[1];
@@ -87,7 +87,7 @@ struct Colour
     }
 
     /// Return this colour with alpha adjusted.
-    inline Colour WithAlpha(float alpha)
+    inline Colour WithAlpha(const float alpha)
     {
         return Colour(r,g,b,alpha);
     }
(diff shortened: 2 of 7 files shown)


christian-rauch:
I merged the creation of the Wayland window and the EGL context into one class, and in the process also solved the blocking `eglSwapBuffers`. The missing `wl_shell_surface_set_toplevel(wshell_surface);` caused blocking when waiting for new events.
In the X11 window interface there is the method for context switching ([`X11Window::MakeCurrent(GLXContext ctx)`](https://github.com/stevenlovegrove/Pangolin/blob/c2a6ef524401945b493f14f8b5b8aa76cc7d71a9/src/display/device/display_x11.cpp#L313)), whose Wayland counterpart has now been removed. Is this method actually used? I haven't found references in Pangolin.

Regarding the issue with glew<2, most examples seem to run fine without glew. Can you give examples where glew is used? Would it be possible to have a flag to de-/activate glew? There is the flag [`HAVE_GLEW`](https://github.com/stevenlovegrove/Pangolin/blob/c2a6ef524401945b493f14f8b5b8aa76cc7d71a9/src/CMakeLists.txt#L168), but it is only used at some places. Or would you alternatively, just make the workaround `FALLBACK_X11_IF_GLEW1` permanent?

christian-rauch:
I added client side decorations for resizing, maximising and closing:
simpledisplay_wayland

There is still an issue with the alpha channel of colours. E.g. in SimplePlot, marking an area with the mouse uses a colour with alpha smaller than 1 and hence shows the background:
simpleplot_wayland

@stevenlovegrove Do you have some comments regarding the `glewInit` crash? Should the `FALLBACK_X11_IF_GLEW1` workaround be permanent?

stevenlovegrove:
Regarding GLEW, I think I have included it mainly for convenience in host applications, since it must be included before other GL headers and it is so common. Thinking about it, this isn't a great reason and should probably be removed, but in the mean-time I think your X11 fallback solution is fine.

Regarding the extra dependencies, I would not like Wayland to be required as you currently have it. On Linux, both X11 and Wayland should be optional, with a check that at least one is found. The fallback logic and default can be applied here:
https://github.com/stevenlovegrove/Pangolin/blob/master/src/display/display.cpp#L139

Regarding Window decorations - how does Wayland usually support this? It seems a bit crazy that this would require custom decoration?

Thanks again for your PR - I'm super excited to get this merged!

christian-rauch:
Thanks for your feedback.
I will then only build the Wayland plugin, if the required libraries are found. Pangolin will then try to load the Wayland library first, and if it fails to connect to a Wayland server/compositor it will fallback to X11.
However, is it ok to keep the dependencies for Travis CI in place (to check that Wayland + X11 builds) or do you also want to have them removed for CI?

Regarding the window decorations: There seems to be no rule about who needs to draw the window decorations. There is a protocol worked on that allows to negotiate with the Wayland server/compositor (i.e. GNOME, KDE, ...) if a Wayland client can use server side decorations (SSD) or if it wants to draw its own decorations (CSD). Sadly, GNOME doesn't support this protocol and a Wayland client needs to draw its own decorations anyway. The usual response from them is that clients should use a toolkit like Qt or Gtk.

christian-rauch:
I reordered the commits to move Wayland independent parts to the beginning, in case you want to merge these first. Otherwise, I fixed the transparent background bug and some other issues with the resizing.

The build and run behaviour is now as proposed: The Wayland plugin is built, if the `wayland-client` libraries are available (by default on Ubuntu 18.04). If an unsupported GLEW version is detected or no Walyand server is available, Pangolin will fall back to X11.

PR #349, opened 2018-05-02
Use target-based include_directories to enable modern CMake usage

stotko:
This PR adds support for modern target-based linking where the include directories are automatically propagated without further commands. The old fashioned way of linking (with `(target_)include_directories`) is still supported for users relying on it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.6)
+cmake_minimum_required(VERSION 2.8.12)
 project("Pangolin")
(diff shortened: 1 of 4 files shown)


stotko:
Modern linking is only enabled on Linux/MacOS and disabled for Windows since GLEW is built externally and its headers are not installed and remain inside the build directory.

stevenlovegrove:
This is great! The Glew / JPEG / PNG builds are not *just* for Windows, but they are default on that platform. Perhaps we could create import targets for those so that this would work there too?

Issue #326, opened 2018-01-26
CMake Error at test/log/CMakeLists.txt:2


geoffviola:
I got the following error while following the instructions on a new computer

```
CMake Error at test/log/CMakeLists.txt:2 (find_package):
  By not providing "FindPangolin.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "Pangolin",
  but CMake did not find one.

  Could not find a package configuration file provided by "Pangolin"
  (requested version 0.4) with any of the following names:

    PangolinConfig.cmake
    pangolin-config.cmake

  Add the installation prefix of "Pangolin" to CMAKE_PREFIX_PATH or set
  "Pangolin_DIR" to a directory containing one of the above files.  If
  "Pangolin" provides a separate development package or SDK, be sure it has
  been installed.
```

It looks like when doing a fresh build and running cmake without parameters, the tests try to run `find_package(Pangolin` without the module being added to the path. A workaround would be to pass `-DBUILD_TESTS=false` to cmake.

ahmadyan:
I have the same issue installing Pangolin on a new mac as well. 
I confirm the work-around.

stevenlovegrove:
The Pangolin package should be exported automatically via the CMake package registry. The CMake exports can be found in ~/.cmake/packages/Pangolin/ on Linux / OSX.

I've had people comment on this problem previously when running without permission to write to that folder (from within Docker or something, for instance). Could this be the case here?

geoffviola:
I don't think it's a permissions issue. I think I could reproduce it in a virtual environment, but that would take some effort. You won't see the issue, unless it's on a machine without Pangolin installed.

geoffviola:
The file looks good. I didn't realize that file got generated on CMake. I wasn't able to reproduce the problem. The problem might be on my install.

Here are the steps that I tried. 
```
docker run -it ubuntu
# In container
cd /home
mkdir temp
cd temp
apt-get update
apt-get install -y git cmake build-essential mesa-common-dev libpng-dev libjpeg-dev libxxf86vm1 libxxf86vm-dev libxi-dev libxrandr-dev freeglut3-dev libglapi-mesa freeglut3-dev
git clone https://github.com/stevenlovegrove/Pangolin.git
cd Pangolin
mkdir build
cd build
cmake ..
# no errors...
```

Since those tests are never compiled, it might be prudent to delete those CMakeLists.txt.

[...]

Issue #286, opened 2017-08-23
cmake error


JingruiYu:
Hi, I am not sure why, but here is a warning and an error:

CMake Warning at src/CMakeLists.txt:652 (export):
  Cannot create package registry file:

    /home/yujr/.cmake/packages/Pangolin/be7d1796df240480052067a613bde09e

  No such file or directory

CMake Error at test/log/CMakeLists.txt:2 (find_package):
  By not providing "FindPangolin.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "Pangolin",
  but CMake did not find one.

  Could not find a package configuration file provided by "Pangolin"
  (requested version 0.4) with any of the following names:

    PangolinConfig.cmake
    pangolin-config.cmake

  Add the installation prefix of "Pangolin" to CMAKE_PREFIX_PATH or set
  "Pangolin_DIR" to a directory containing one of the above files.  If
  "Pangolin" provides a separate development package or SDK, be sure it has
  been installed.

-- Configuring incomplete, errors occurred!

could you help me?

sajjadt:
A workaround that worked for me is to disable tests and examples. 
```
cmake -DBUILD_TESTS=OFF -DBUILD_EXAMPLES=OFF ..
```

NikolausDemmel:
See also #326.

As commented in https://github.com/stevenlovegrove/Pangolin/issues/326#issuecomment-392140263, I think it would be best if tests and examples would not rely on the package registry to ensure they always find the correct version of pangolin when you have multiple builds from different projects on the same system.

PR #358, opened 2018-05-27
Set Pangolin_DIR for tests. Fixes #357, #326, #286.

### LINKED ISSUE #357: Better story for EXPORT_Pangolin=Off
I think it's worth to open a separate issue for this. I previously commented about this: https://github.com/stevenlovegrove/Pangolin/issues/326#issuecomment-392140263 and https://github.com/stevenlovegrove/Pangolin/issues/286#issuecomment-392143818.

If you build pangolin without the library export to the package registry at `~/.cmake/packages/...` with

```
cmake .. -DEXPORT_Pangolin=Off
```

then the examples and tests don't build 

```
CMake Error at test/log/CMakeLists.txt:2 (find_package):
  By not providing "FindPangolin.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "Pangolin",
  but CMake did not find one.

  Could not find a package configuration file provided by "Pangolin"
  (requested version 0.4) with any of the following names:

    PangolinConfig.cmake
    pangolin-config.cmake

  Add the installation prefix of "Pangolin" to CMAKE_PREFIX_PATH or set
  "Pangolin_DIR" to a directory containing one of the above files.  If
  "Pangolin" provides a separate development package or SDK, be sure it has
  been installed.
```

You have to explicitly and also may not enable tests
```
cmake .. -DEXPORT_Pangolin=Off -DBUILD_TESTS=Off -DBUILD_EXAMPLES=Off
```

Maybe it would be good if cmake would check `EXPORT_Pangolin` before trying to compile tests and example. But probably it would be better if the tests and examples would not rely on the global package registry, but make sure to always find the pangolin lib from the same build tree. The current approach seems error prone if you have multiple copies of pangolin in different projects compiled at the same time, since it might pick up the wrong one.

For example, this might be caused by this (only speculation): #307

NikolausDemmel:
Fixes #357, #326, #286.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -94,6 +94,7 @@ add_subdirectory("external")
 add_subdirectory("src")
 
 if(BUILD_TESTS)
+  set(Pangolin_DIR ${Pangolin_BINARY_DIR}/src)
   add_subdirectory("test")
 endif()
 


stevenlovegrove:
Thanks!

Issue #363, opened 2018-06-03
comment out pointers


rpfly3:
@stevenlovegrove Thanks for the excellent code. Here is a issue I found for the latest code.
`
//static void pointer_handle_frame(void */*data*/, struct wl_pointer */*wl_pointer*/) { }

//static void pointer_handle_axis_source(void */*data*/, struct wl_pointer */*wl_pointer*/, uint32_t /*axis_source*/) { }

//static void pointer_handle_axis_stop(void */*data*/, struct wl_pointer */*wl_pointer*/, uint32_t /*time*/, uint32_t /*axis*/) { }

//static void pointer_handle_axis_discrete(void */*data*/, struct wl_pointer */*wl_pointer*/, uint32_t /*axis*/, int32_t /*discrete*/) { }

static const struct wl_pointer_listener pointer_listener = {
    pointer_handle_enter,
    pointer_handle_leave,
    pointer_handle_motion,
    pointer_handle_button,
    pointer_handle_axis,
#if WAYLAND_VERSION_GE(1,12)
    pointer_handle_frame,
    pointer_handle_axis_source,
    pointer_handle_axis_stop,
    pointer_handle_axis_discrete,
#endif
};
`

Why leave pointers like handle_frame etc there? There are some compile problems.

kzampog:
This is the same as https://github.com/stevenlovegrove/Pangolin/issues/364.

Issue #364, opened 2018-06-03
Problem when making on ubuntu 16.04


Cheren15:
I have came across the problem when I making as follow:

src/CMakeFiles/pangolin.dir/build.make:1574: recipe for target 'src/CMakeFiles/pangolin.dir/display/device/display_wayland.cpp.o' failed
make[2]: *** [src/CMakeFiles/pangolin.dir/display/device/display_wayland.cpp.o] Error 1
CMakeFiles/Makefile2:136: recipe for target 'src/CMakeFiles/pangolin.dir/all' failed
make[1]: *** [src/CMakeFiles/pangolin.dir/all] Error 2
Makefile:149: recipe for target 'all' failed
make: *** [all] Error 2

And I have installed the Glew and tried altered the PYTHON thing in ccmake, it would not work for me. 
So I directly turn off the pangolin GUI in ccmake and bypass this module to build, I was wondered if this will affect future using and is there any better way to solve the problem??

kzampog:
I am also unable to build `display_wayland.cpp` on Mint 18.3:

```
/home/kzampog/Coding/Libraries/Pangolin/src/display/device/display_wayland.cpp:600:5: error: ‘pointer_handle_frame’ was not declared in this scope
     pointer_handle_frame,
     ^
/home/kzampog/Coding/Libraries/Pangolin/src/display/device/display_wayland.cpp:601:5: error: ‘pointer_handle_axis_source’ was not declared in this scope
     pointer_handle_axis_source,
     ^
/home/kzampog/Coding/Libraries/Pangolin/src/display/device/display_wayland.cpp:602:5: error: ‘pointer_handle_axis_stop’ was not declared in this scope
     pointer_handle_axis_stop,
     ^
/home/kzampog/Coding/Libraries/Pangolin/src/display/device/display_wayland.cpp:603:5: error: ‘pointer_handle_axis_discrete’ was not declared in this scope
     pointer_handle_axis_discrete,
     ^
```

zxp771:
Hi @kzampog
I got the same problem.
I also  modified the display_wayland.cpp as d7b6942 
but it still not work.
Do you have any other method?

kzampog:
I just uncommented lines  585-591 of `display_wayland.cpp` and it compiled fine on my Mint 18.3!

PeterQFR:
you can use commit f499c390b282 on master branch works. Seems some recent changes broke it.

zxp771:
[...]

PR #366, opened 2018-06-06
Fix wayland warnings

christian-rauch:
This PR defines the wayland callbacks again if they are supported and used.
Fixes https://github.com/stevenlovegrove/Pangolin/issues/363 and https://github.com/stevenlovegrove/Pangolin/issues/364 but should also silent warnings about unused functions.


stevenlovegrove:
Oops - sorry!

PR #369, opened 2018-06-19
Remove unnecessary stdlib flag, allowing user to override if necessary.

### LINKED ISSUE #356: Compiling with clang on Linux, libc++ vs libstdc++
When I compile Pangolin on Linux with clang, it compiles with `libc++`: https://github.com/stevenlovegrove/Pangolin/blob/4335035057dc26f98b1a498b9b4843015fc9e5c5/CMakeLists.txt#L53

This causes issue when later using it in another project which is compiled with the default `libstdc++`. You get errors like:

```
undefined reference to `pangolin::CreateWindowAndBind(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, int, int, pangolin::Params const&)'
```

If that particular line in the `CMakeLists.txt` is removed, clang defaults to `libstdc++` on Linux and everything works fine.

Do you remember why that line is there? It looks like it has been for 5 or more years. Maybe back then things were different, but I believe that clang defaults to the native STL implementation on each platform, i.e. to `libstdc++` on Linux and `libc++` on macOS. So maybe that line can be removed?

NikolausDemmel:
Fix #356.

I added `libstdc++-5-dev` from the `ubuntu-toolchain-r-test` PPA to make clang work on Linux (looks like the system default for trusty is too old, namely libstdc++-4.8).

The failure of GCC in Linux seems to be from master, but interestingly also fails in a libstdc++-4.8 header it seems.

Note that travis is already on clang-5.0, so no need to set it up manually (as you attempt in https://github.com/stevenlovegrove/Pangolin/commit/b98d8f70ea822de205dada9d2480d471a5fd9f37).

I could add to the README a note for Trusty users that want to use clang to install newer libstdc++ or pass flags to use libc++ is needed. Should I do that?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,14 @@ before_install:
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update          ; fi
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install eigen glew ; fi
 
+addons:
+  apt:
+    sources:
+      - ubuntu-toolchain-r-test
+    packages:
+      # install newer libstdc++ (from ubuntu-toolchain-r/test PPA), since otherwise clang fails with old libstdc++-4.8
+      - libstdc++-5-dev
+
 language: cpp
 
 compiler:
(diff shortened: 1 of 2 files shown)


NikolausDemmel:
Looks like this is ready to merge.

PR #380, opened 2018-06-28
Fix min/max values for Attach with log scale.

NikolausDemmel:
Without this the min-max values are interpreted as powers of `e`.

The code section is copied from the corresponding `Var` constructor further down.

Tested on macOS.


stevenlovegrove:
Thanks!

PR #386, opened 2018-07-13
Test scenehandler with follow

lefutonku-github:
Recently I am trying to use the scenehandler to facilitate an selectable pangolin window, like the following picture. the original cubes in the window are green, i want them to recognize the mouse 'click' on them and then change to red color.

image

    I use the sceneHandler which seems designed for this scenarios, it works fine in the beginning, but after i use OpenGlRenderState::Follow(mat) in the example,  the cube immediately become un-recognizable, which is caused by the  always zero value returned in SceneHandler::ComputeHits.

    I did some debug in GDB and in my understanding, it's seems caused by the function here
image

    One the OpenGlRenderState is set to follow an un-indentical matrix, the 'T_cw' and 'modelview' field will change accordingly, then here comes the difference between GL_RENDER mode and GL_SELECT mode, so it's unable to hit anything. see:
1. the GL_RENDER mode, the model view matrix is multiplied by T_CW
image
2. the GL_SELECT mode, not multiplied with T_CW
image

    i just raise this very messy pull request for understanding the problem and propose the possible fix. a more cleaner pull request can be raised once it's confirmed


stevenlovegrove:
Thanks for this and my apologies for not getting back to you sooner. I will try to test this out soon.

Issue #393, opened 2018-07-30
Using shared context in mutliple threads


christian-rauch:
I am trying to use Pangolin (X11+GLX) from multiple threads, i.e. I am creating a window in the main thread and drawing into this window from another thread.

Here is the modified HelloPangolin example, using a thread to draw into the window:

```
#include <pangolin/pangolin.h>
#include <thread>
#include <pangolin/display/display_internal.h>

void setup() {
    pangolin::CreateWindowAndBind("Main",640,480);
    glEnable(GL_DEPTH_TEST);

    std::cout << "current ctx setup: " << pangolin::GetCurrentContext() << std::endl;
}

void run() {
    std::cout << "current ctx run: " << pangolin::GetCurrentContext() << std::endl;

//    pangolin::BindToContext("Main");    // <- this blocks
    pangolin::FindContext("Main")->MakeCurrent();

    std::cout << "current ctx run: " << pangolin::GetCurrentContext() << std::endl;

    // Define Projection and initial ModelView matrix
    pangolin::OpenGlRenderState s_cam(
        pangolin::ProjectionMatrix(640,480,420,420,320,240,0.2,100),
        pangolin::ModelViewLookAt(-2,2,-2, 0,0,0, pangolin::AxisY)
    );

    // Create Interactive View in window
    pangolin::Handler3D handler(s_cam);
    pangolin::View& d_cam = pangolin::CreateDisplay()
            .SetBounds(0.0, 1.0, 0.0, 1.0, -640.0f/480.0f)
            .SetHandler(&handler);

    while( !pangolin::ShouldQuit() )
    {
        // Clear screen and activate view to render into
        glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
        d_cam.Activate(s_cam);

        // Render OpenGL Cube
        pangolin::glDrawColouredCube();

        // Swap frames and Process Events
        pangolin::FinishFrame();
    }
}

int main( int /*argc*/, char** /*argv*/ )
{
    setup();

    std::thread render_loop;
    render_loop = std::thread(run);
    render_loop.join();

    return 0;
}
```
Getting the context via `pangolin::BindToContext("Main")` blocks the execution of `run()`. `pangolin::FindContext("Main")->MakeCurrent()` returns the same context, but the window is empty (black). Calling `run()` outside of a thread correctly draws the cube.

What is the right way to use Pangolin in multiple threads?

[...]

Issue #388, opened 2018-07-23
Pangolin X11: Unable to retrieve framebuffer options


christian-rauch:
I am also getting the `Pangolin X11: Unable to retrieve framebuffer options` when calling `pangolin::CreateWindowAndBind("lala", 10, 10);`. In contrast to https://github.com/stevenlovegrove/Pangolin/issues/242 and https://github.com/stevenlovegrove/Pangolin/issues/260, I am running natively under Linux X11.

However, this error only occurs when creating the window inside a class (let's say class A) method. It does not appear when creating the window in the class A constructor but I get errors when calling pangolin methods in class A methods.

Are there limitations how pangolin can be used in different threads?

The documentation of [glXChooseFBConfig](https://www.khronos.org/registry/OpenGL-Refpages/gl2.1/xhtml/glXChooseFBConfig.xml) mentiones:
`NULL is returned if an undefined GLX attribute is encountered in attrib_list, if screen is invalid, or if dpy does not support the GLX extension.`

Is there a way to test which attributes in `attrib_list` are unsupported and which GLX extension is unsupported? Any ideas why these settings might change in between the two cases mentioned earlier?

christian-rauch:
I started to replace GLX by EGL for the X11 window: https://github.com/stevenlovegrove/Pangolin/pull/389 (https://github.com/christian-rauch/Pangolin/tree/x11_egl). This circumvents the problem with GLX frame buffers.

christian-rauch:
The error I get when calling pangolin functions from the class A method seem to be related to the `context`.

When looking at the pointers of the window and context via:
```C++
std::cout << "win: " << pangolin::GetBoundWindow() << std::endl;
std::cout << "ctx: " << pangolin::GetCurrentContext() << std::endl;
std::cout << "find ctx: " << pangolin::FindContext(window_name) << std::endl;
```
I get the same addresses when called in the class A constructor after `CreateWindowAndBind()`:
```
win: 0x7fae37996070
ctx: 0x7fae37996070
find ctx: 0x7fae37996070
```
But when called in the class A method, I get:
```
win: 0
ctx: 0
find ctx: 0x7fae37996070
```
which suggests that I need to set the current context to the stored context. Is this because of the original context going out of scope after the constructor?

Setting the current context in the class A method via:
```C++
pangolin::BindToContext(window_name);
```
blocks, and doing it via:
```C++
pangolin::FindContext(window_name)->MakeCurrent();
```
[...]

Issue #374, opened 2018-06-21
window-less offscreen rendering


christian-rauch:
Is it possible to use Pangolin without a window, i.e. for pure offscreen rendering?

I am doing  off-screen rendering using `pangolin::GlFramebuffer` and a `glViewport` to render an image that might be larger than the screen size. I can than access the pixel values via `glReadPixels`. To setup pangolin, you need to call `CreateWindowAndBind` even if you do not draw into this window.

I guess that since the X11 window is using GLX, there is no way around rendering without a window. I have seen that via EGL, it's possible to render without an X-server at all: https://devblogs.nvidia.com/egl-eye-opengl-visualization-without-x-server/. Would using EGL for X11 help in this situation? Could EGL also be used in place of WGL (Windows) and CGL (Mac OS X)?

stevenlovegrove:
Pangolin does not currently support that, but yes, implementing an EGL context is probably the best way to go. With the new Window factory, I think it would be good to offer it as a window URI, too.

christian-rauch:
This is now implemented in https://github.com/stevenlovegrove/Pangolin/commit/a69487917cca852cd90c883cb94795f92316bc7a.

PR #394, opened 2018-08-04
GLFW window and context creation

christian-rauch:
This PR adds GLFW for window and context creation as an alternative to manually creating the different contexts on the different operating systems. GLFW is basically a abstraction on top of the native windows, X11, wayland and mac os calls.
At the moment the glfw upstream project is simply added as git submodule and linked statically to pangolin.

The PR mainly solves the issues:
- EGL context creation (e.g. prevents GLX issue https://github.com/stevenlovegrove/Pangolin/issues/388)
- offscreen rendering (issue https://github.com/stevenlovegrove/Pangolin/issues/374)
- context sharing in different threads (issue https://github.com/stevenlovegrove/Pangolin/issues/393)

GLFW requires CMake > 3.0 (i.e. it will not support Ubuntu 14.04) and `Pangolin_FOUND` is currently not set because of:
```
  Found package configuration file:

    [...]/PangolinConfig.cmake

  but it set Pangolin_FOUND to FALSE so package "Pangolin" is considered to
  be NOT FOUND.  Reason given by package:

  The following imported targets are referenced, but are missing: glfw
```
I.e. something like `find_package(Pangolin 0.5 REQUIRED)` will fail but `find_package(Pangolin 0.5)` will work.

Any ideas how to add glfw as an import target to pangolin? My intention with static linking was actually to prevent these kind of issues.


christian-rauch:
I added example SimpleMultiDisplayOffscreen to demonstrate the offscreen rendering. The offscreen rendering doe snot create a window, but it still requires a running xserver 
It does not create a window, but it still requires a running Xorg server.

stevenlovegrove:
Interesting - I can't say I'm keen on adding GLFW as a required dependency but maybe as an optional one. I guess your other PR is for EGL on Linux only and hence offscreen rendering on that platform only?

christian-rauch:
Yes, I looked into GLFW as an alternative for making X11+EGL work and to get offscreen rendering with a simple flag. It's not of high priority for me at the moment. I wouldn't mind if it gets rejected.
However, in the longterm I think it's easier to just maintain GLFW in Pangolin than all the other OS specific bindings (Windows, X11, Wayland, macOS). E.g., if macOS really drops OpenGL support at some point, relying on a third party library will make things easier.

Issue #355, opened 2018-05-26
Using Pangolin in cmake project with custom Eigen version


NikolausDemmel:
I would like to use Pangolin in another cmake project, including its source tree as a git submodule. However, I am no cmake expert, and thus I'm asking for advice, experience, best-practices.

One simple way would be to just compile Pangolin beforehand, letting cmake's package registry in `~/.cmake/packages/...` ensure that a subsequent `find_package(Pangolin)` call in my projects cmake source finds the location. But as hinted in https://github.com/stevenlovegrove/Pangolin/issues/326#issuecomment-392140263, I've had trouble before with this approach, where cmake would pick up the "wrong" pangolin build tree, for example if I have multiple projects coming with Pangolin, or when I have different builds for debug / release / etc.

I think ideally I would want to include Pangolins sources directly in my project with `add_subdirectory` to directly depend on it's targets and get best build-parallelization. I guess this would also be best if sometimes I need to tweak and change Pangolin's source. However, it seems this would require a lot of fiddling with Pangolin's cmake. As it is, if I do this, Pangolin's top-level cmake creates conflicting targets like `uninstall` or `doc`. I guess it is simply not designed to be used that way. Is anyone successfully using Pangolin with `add_subdirectory`?

An alternative that is probably a bit more robust to the specifics of Pangolins cmake might be cmake's `ExternalProject`. One would need to ensure that all the needed parameters are passed to the cmake sub-process, also making sure pangolin does not export itself to the package registry. When using pangolin the parent project's cmake, one could either do `find_package`, specifying the exact location where it should search for the external-project to ensure it does not find another Pangolin build / install tree. It seems like pangolin itself uses external projects slightly differently in https://github.com/stevenlovegrove/Pangolin/blob/master/external/CMakeLists.txt (although that seems to be specific to windows in that instance). But I guess a similar thing could be done, creating custom targets and setting `Pangolin_INCLUDE_DIRS` and `Pangolin_LIBRARIES` appropriately. Does anyone have an example of using Pangolin in that way?

I'd appreciate thoughts on what is the best approach and/or pointers to examples.

stevenlovegrove:
Pangolin is designed to support this use case, though I haven't tried it recently. You can see an example in this project: https://github.com/arpg/CoreDev

If there are issues (like duplicate target names), I'm happy to help you resolve them - i.e. by renaming docs to pangodocs or similar

stevenlovegrove:
Notice the line:
[...]

PR #419, opened 2018-10-16
Use unique uninstall & doc targets

powderluv:
When Pangolin is used in other projects these targets collide

Also fixes https://github.com/stevenlovegrove/Pangolin/issues/355 

TEST: Build on OSX and run samples

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -635,12 +635,12 @@ if(BUILD_PYPANGOLIN_MODULE AND HAVE_PYTHON )
 endif()
 
 #######################################################
-## Generate Doxygen documentation target (make doc)
+## Generate Doxygen documentation target (make pangolin_doc)
 find_package(Doxygen)
 if(DOXYGEN_FOUND)
     configure_file(${CMAKE_CURRENT_SOURCE_DIR}/Doxyfile.in ${CMAKE_CURRENT_BINARY_DIR}/Doxyfile @ONLY)
     file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/../doc )
-    add_custom_target(doc
+    add_custom_target(pangolin_doc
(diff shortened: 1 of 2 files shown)


src/CMakeLists.txt
```diff
@@ -640,7 +640,7 @@ find_package(Doxygen)
 if(DOXYGEN_FOUND)
     configure_file(${CMAKE_CURRENT_SOURCE_DIR}/Doxyfile.in ${CMAKE_CURRENT_BINARY_DIR}/Doxyfile @ONLY)
     file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/../doc )
-    add_custom_target(doc
+    add_custom_target(pangolin_doc
```
NikolausDemmel: You should also update the docstring above that currently reads `(make doc)`

NikolausDemmel:
Thanks for this PR. I think these changes should be merged right away.

> Also fixes #355

However, to address all points in #355 one would also have to resolve the issue when using this with a specified Eigen version that is internal to the parent project (i.e. within the source directory of the parent project), so I don't think this PR should close #355.

src/CMakeLists.txt
powderluv: done

powderluv:
Ok on not closing #355. I have a few more patches to help build on windows / with GLEW variations etc  that I will clean up from https://github.com/powderluv/Pangolin and submit PRs.

NikolausDemmel:
Awesome thanks.

:+1: for merging.

stevenlovegrove:
Thanks!

leCaTo:
An updated README with the new target name would be nice:
'cmake --build . --target pangolin_doc' instead of 'cmake --build . --target doc'

NikolausDemmel:
@leCaTo can you send a PR?

Issue #414, opened 2018-10-09
Cmake error on windows 10


BrunoXL:
I got the error below on building `Pangolin` on windows 10. 

     -- Selecting Windows SDK version 10.0.16299.0 to target Windows 10.0.17134.
     CMake Error at C:/Program Files/CMake/share/cmake-3.11/Modules/ExternalProject.cmake:1654 (file):
     file problem creating directory: C:/Program Files (x86)/Pangolin
     Call Stack (most recent call first):
     C:/Program Files/CMake/share/cmake-3.11/Modules/ExternalProject.cmake:3052 (_ep_set_directories)
     external/CMakeLists.txt:10 (ExternalProject_Add)

Work-jk-l:
The same error in my project on window10(vs2015,update3)，Compile the pangolin.sln is successful,but many errors in compile my own project using the genetrated *.lib.。So I am wondering whether pangolin can not be use on windows?

powderluv:
You can try to pass CMAKE_INSTALL_PREFIX to work around that issue. It can't handle spaces in the file path.

powderluv:
try https://github.com/stevenlovegrove/Pangolin/pull/420

Gatunox:
BrunoXL

You need to be administrator, so please excute the cmd or VS2017 as an administrator an it will work, also check the **build on windows** issue, i've provided a detailed explanation to compile on windows.

regards

navalsheth:
Ture 

> For any people encountering weird build error on Windows 10,
> 
> please make sure you run CMake-GUI and Visual Studio in Administrator mode !
> 
> Within administrator mode, the installation should be easy.

TRUE, I got stuck for one week with same error as mentioned above and then I Opened CMAKE as Administrator, Then things worked well

lz2004xxw:
i wonder why it worked in Administrator mode ?

valeriob:
Probably because it's trying to install into C:\Program Files (x86)\ 
i just changed that parameter in cmake and it worked without admin privileges

PR #420, opened 2018-10-16
Fix cmake default

powderluv:
Set up a sane CMAKE install prefix by default to help build on Windows


powderluv:
@stotko Decoupled this patch from #419  I verified this on windows 10 / VS 2017 / CMake 3.12. Without this patch I run into #414  mkdir failing since the default "C:\Program Files.." has a space. With this patch I can get CMake to complete successfully.  The issue is because CMake itself doesn't like spaces when making directories. https://github.com/Kitware/CMake/blob/88ef5737a25b40ce738041b8f9e00df20a131eb6/Modules/ExternalProject.cmake#L1648

"  # Make the directories at CMake configure time *and* add a custom command
  # to make them at build time. They need to exist at makefile generation
  # time for Borland make and wmake so that CMake may generate makefiles
  # with "cd C:\short\paths\with\no\spaces" commands in them."

There are still issues with GLEW cmake dependencies not exposed properly. I am working with upstream GLEW https://github.com/nigels-com/glew/issues/149#issuecomment-428446636 in the hope that they provide good CMake support so we dont have to use a fork.  I will submit PRs for those issues next.

stotko:
Two other comments from my side:
- This seems to be a windows-specific issue so the code should be guarded with `if(_WIN_)`
- Users might not expect this behaviour and be confused. Adding a short message, e.g. `message( STATUS "Changed default install prefix to ${CMAKE_INSTALL_PREFIX}") ` would help here.

powderluv:
@stotko  done

stevenlovegrove:
Do you have a reference to this being a specific CMake limitation as opposed to a Pangolin CMake bug? I tried googling but didn't find anything concrete.

powderluv:
Yes please see comments here: https://github.com/Kitware/CMake/blob/88ef5737a25b40ce738041b8f9e00df20a131eb6/Modules/ExternalProject.cmake#L1648

stevenlovegrove:
That comment is in the ExternalProject module. Do you know that it relates to install targets?

powderluv:
yes it does relate to the install target. See this issue https://github.com/stevenlovegrove/Pangolin/issues/414

The second line the error is because of the default INSTALL prefix having a space. This PR mitigates it. Unless you want to install in a directory which has a space. Then good luck :) 

> CMake Error at C:/Program Files/CMake/share/cmake-3.11/Modules/ExternalProject.cmake:1654 (file):
> file problem creating directory: C:/Program Files (x86)/Pangolin

PR #423, opened 2018-10-17
Pythonic video iterations

martinruenz:
This pull request makes the python binding of `VideoInput` iterable and enables a more pythonic way of running over an input sequence. The according python example is also updated and a minor bug in the example is fixed.
The (broken) python interfaces for the functions `GrabNext` and `GrabNewest` are removed.


stevenlovegrove:
Great! Thanks!

Issue #331, opened 2018-02-14
Linker Problem (visualStudio, Win10)


gaeae:
Visual Studio 15 2017 (v141)
Windows 10
x64
Release and Debug

I proceeded as described in the manual. (Without any problems!)

Now added:
- include directories (/include ; /build/src/include/ and build/external/glew/include)
- library derectories (/build/src/Release/ ; build/external/glew/lib/)
- additional dependencies (pangolin.lib ; glew.lib ; glew_shared.lib)
- copied pangolin.dll and glew.dll in the solution directory

When including pangolin.h (in my Viewer.cpp file) the following error occurs: 

> 1>Viewer.obj : error LNK2001: unresolved external symbol "public: virtual void __cdecl pangolin::Handler::Keyboard(struct pangolin::View &,unsigned char,int,int,bool)" (?Keyboard@Handler@pangolin@@UEAAXAEAUView@2@EHH_N@Z)
> 1>Viewer.obj : error LNK2001: unresolved external symbol "public: virtual void __cdecl pangolin::Handler::Mouse(struct pangolin::View &,enum pangolin::MouseButton,int,int,bool,int)" (?Mouse@Handler@pangolin@@UEAAXAEAUView@2@W4MouseButton@2@HH_NH@Z)
> 1>Viewer.obj : error LNK2001: unresolved external symbol "public: virtual void __cdecl pangolin::Handler::MouseMotion(struct pangolin::View &,int,int,int)" (?MouseMotion@Handler@pangolin@@UEAAXAEAUView@2@HHH@Z)
> 1>Viewer.obj : error LNK2001: unresolved external symbol "public: virtual void __cdecl pangolin::Handler::PassiveMouseMotion(struct pangolin::View &,int,int,int)" (?PassiveMouseMotion@Handler@pangolin@@UEAAXAEAUView@2@HHH@Z)
> 1>Viewer.obj : error LNK2001: unresolved external symbol "public: virtual void __cdecl pangolin::Handler::Special(struct pangolin::View &,enum pangolin::InputSpecial,float,float,float,float,float,float,int)" (?Special@Handler@pangolin@@UEAAXAEAUView@2@W4InputSpecial@2@MMMMMMH@Z)
[...]

PR #424, opened 2018-10-17
Import external dependencies as git submodules

powderluv:
This helps avoid the pesky External_ProjectAdd in CMake
Also disable examples if we use BUILD_EXTERN_GLEW etc since
tests assume all the packages are installed.

TEST:builds libpangolin.${SUFFIX} on
1) Windows 10/VS 2017
2) OSX 10.14 Mojave
3) Ubuntu 18.04

all three OSes are built with the following command

"cmake .. -GNinja -DBUILD_EXTERN_GLEW=ON; ninja"

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -87,7 +87,6 @@ configure_file(
 add_custom_target(uninstall
   "${CMAKE_COMMAND}" -P "${CMAKE_CURRENT_BINARY_DIR}/cmake_uninstall.cmake")
 
-add_subdirectory("external")
 add_subdirectory("src")
 
 if(BUILD_TESTS)
@@ -95,7 +94,8 @@ if(BUILD_TESTS)
   add_subdirectory("test")
 endif()
 
-if(BUILD_EXAMPLES)
+if(BUILD_EXAMPLES AND NOT BUILD_EXTERN_GLEW)
+  #if we build glew you will need to update the examples to point to the GLEW headers/libs
   set(Pangolin_DIR ${Pangolin_BINARY_DIR}/src)
   add_subdirectory(examples)
   add_subdirectory(tools)
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,15 @@
 [submodule "external/pybind11"]
 	path = external/pybind11
 	url = https://github.com/pybind/pybind11.git
+[submodule "external/glew"]
+	path = external/glew
+	url = https://github.com/Perlmint/glew-cmake.git
+[submodule "/external/zlib"]
+	path = external/zlib
+	url = https://github.com/madler/zlib.git
+[submodule "external/libpng"]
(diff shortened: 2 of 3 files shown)


CMakeLists.txt
```diff
@@ -87,15 +87,15 @@ configure_file(
...
-add_subdirectory("external")
 add_subdirectory("src")
 
 if(BUILD_TESTS)
   set(Pangolin_DIR ${Pangolin_BINARY_DIR}/src)
   add_subdirectory("test")
 endif()
 
-if(BUILD_EXAMPLES)
+if(BUILD_EXAMPLES AND NOT BUILD_EXTERN_GLEW)
```
stevenlovegrove: The user may have built GLEW elsewhere - GLEW is actually listed as a required dependency.

stevenlovegrove:
Thanks for this PR. One reason I went with External_ProjectAdd was because I didn't want users on Linux or OSX to have to download or build these libs because they are trivial to get from a package manager. I also wanted for people who downloaded the zip instead of using git (which seems to be quite common for Windows users or those without a github account) to be able to get these packages automatically. Have you had a particular issue with the External_ProjectAdd method?

powderluv:
I ran into multiple issues: (You will notice a repeat of number of people unable to build on Windows).

1) External_ProjectAdd() fails if CMAKE_INSTALL_PREFIX has a space (like the default on  Windows).
https://github.com/Kitware/CMake/blob/88ef5737a25b40ce738041b8f9e00df20a131eb6/Modules/ExternalProject.cmake#L1648
Appveyor uses an install prefix without a space so it works there. 
https://github.com/stevenlovegrove/Pangolin/issues/425
https://github.com/stevenlovegrove/Pangolin/issues/414
https://github.com/stevenlovegrove/Pangolin/issues/331

2) By products of the external libs are not exposed to the top level CMake when using Ninja. This is only relevant if you use Ninja but with Visual Studio now supporting CMake and Ninja it is the cross platform choice. You can work around it by using the BYPRODUCTS but you need to call out the byproduct.

3) All the projects in external/CMakelists.txt assume _WIN_ now. So .lib is hardcoded. I went down the path of fixing that up to support OSX/Linux. Not to mention glew-cmake decided to add a "d" to debug libs.  I am trying to get glew upstream to support CMake as a first class build system and not rely on hand coded Makefiles. 

Overall I think Pangolin has been so awesome to use cross platform but between its CMake and GLEW's CMake it makes it very hard to build across all platforms if you are including it as part of a larger CMake project. 

I think Pangolin will benefit greatly in reducing variations of what is installed and what is built. If someone wants pre-builts you have them.  If someone wants to build from source I think we shouldn't optimize for ease of install via apt-get / brew and just build all dependencies from source always. Tensorflow is a very good example of this (though I would have preferred CMake be the tool of choice).

PR #429, opened 2018-10-24
Scroll in LayoutVertical with not shown displays

caselitz:
Realizing the scrolling by using the show member of a view causes conflicts when the user also wants to show/hide a display. That's why I introduced a dedicated member scroll_show instead.

I tested it and it works fine for my purposes. As I'm not very familiar with the internals of Pangolin please double check.

### DIFF
--- a/src/plot/plotter.cpp
+++ b/src/plot/plotter.cpp
@@ -829,6 +829,8 @@ void Plotter::SetView(const XYRangef& range)
 
     px.rview.x = range.x;
     py.rview.y = range.y;
+    px.target.x = range.x;
+    py.target.y = range.y;
 }
 
 void Plotter::SetViewSmooth(const XYRangef &range)


caselitz:
Can anyone review this PR? Any chance to get this merged @stevenlovegrove?

stevenlovegrove:
Sorry for not getting back to you sooner - I don't think I quite understand what you're trying to accomplish? Could you elaborate?

caselitz:
Thanks for your reply, I'll try to make it more clear:

I have a Panel with LayoutVertical to which I add multiple sub-Displays. The height of them adds up so that it is necessary to scroll through the Panel to reach all of them.

Now, I ("the user") want to be able to show/hide individual sub-Displays depending on external conditions. I think the `show` property of a panel (inherited from View) is intended to provide this functionality.

Unfortunately, the _scroll_ functionality is also realized by using this `show` property (to determine whether a View needs to be shown depending on the scroll). The issue is that using the `show` property for two different purposes creates conflicts.

Example: consider a shown/visible sub-Display, now hide it (user request), then scroll down so that it needs to be hidden for scroll purposes, scroll up again. The sub-Dispay will be shown again cause the scroll up action set  `show` back to true even though the user wants it to be hidden.

To resolve the conflict, I introduced a dedicated variable `scroll_show` that only represents whether or not a View needs to be shown depending on scrolling. This way, the independent `show` variable can explicitly serve the purpose of showing/hiding a View based on user request.

Issue #367, opened 2018-06-07
Strange problems on using Pangolin with MSVC


WsbPro:
Hi, I compile Pangolin on MSVC2017 (Windows 10 1803) in 64 bit. The compile process finished successfully, but the examples compiled at the same time cannot work as it should be.

![untitled1](https://user-images.githubusercontent.com/5794638/41075461-50c6c5ee-6a3f-11e8-81ec-744a1bfe4384.png)

The result for using the "SimpleDisplay" example is shown in the picture above, the result is obviously incorrect; however, if we change the size of the window in a small bit, the correct result will appear:

![untitled2](https://user-images.githubusercontent.com/5794638/41075494-71ec9c3a-6a3f-11e8-9216-ead01ebc1c80.png)

Any ways of fixing this?

wangniancai:
same  on vs 2015 update3

stevenlovegrove:
Yes - this issue has been introduced recently in some refactoring of the Window logic. I don't have a Windows machine setup right now to debug. Any help would be appreciated!

wangniancai:
@stevenlovegrove  I'm not quite familiar with opengl ,could you please tell me which version work well on windows recently  ,Maybe I can debug this

stevenlovegrove:
The last commit that likely worked is 33b4d430491c39105629f3031f403399e6206534

2cbc1b4da0332c7233438895b9a60833e0b57606 is probably the commit that broke things on Windows.

In particular, this line got removed
https://github.com/stevenlovegrove/Pangolin/commit/2cbc1b4da0332c7233438895b9a60833e0b57606#diff-c5e46dd9b58633c342468ece85e5809bL464

I don't know the non-hacky way to fix this! This is less an OpenGL problem and more of a Windows windowing problem, I think.

hkang123:
display_win.cpp , line 194, 195

I changed the values.... It works...

WinWindow::WinWindow(
    const std::string& window_title, int width, int height
) : hWnd(0)
{
    const HMODULE hCurrentInst = GetModuleHandle(0);
    RegisterThisClass(hCurrentInst);

	PangolinGl::windowed_size[0] = width;// 0;
	PangolinGl::windowed_size[1] = height;// 0;

stevenlovegrove:
@hkang123, are you able to make a pull request? Thanks!

Work-jk-l:
[...]

PR #436, opened 2018-11-01
Fix Windows black screen

powderluv:
Initialize the default Windows windows to width/height instead
of 0.

Fixes: https://github.com/stevenlovegrove/Pangolin/issues/367

TEST: Runs the windows samples without having to resize


stevenlovegrove:
Thank you, thank you!

PR #443, opened 2018-11-13
Headless windows via EGL

christian-rauch:
This PR implements offscreen rendering via EGL's pixel buffer surface (fixes issue https://github.com/stevenlovegrove/Pangolin/issues/374). The new `HeadlessWindow` is defining the scheme `headless` that needs to be set either via an environment variable, e.g.: `export PANGOLIN_WINDOW_URI=headless:`, or via a new scheme parameter for `CreateWindowAndBind`, as shown in the `HelloPangolinOffscreen` example.

Since this feature is using EGL, it currently suffers from the same nvidia driver issues discussed in https://github.com/stevenlovegrove/Pangolin/pull/389. That is, on some combinations of Linux distribution and nvidia driver, the screen is black / rendered image is empty. E.g. Ubuntu 16.04 gives a blank screen/image and the newer Ubuntu 18.04 correctly renders via EGL.
However, this PR does not effect the other window and context creation routines.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -1,6 +1,7 @@
 # All examples depend on Pangolin GUI
 if(BUILD_PANGOLIN_GUI)
     add_subdirectory(HelloPangolin)
+    add_subdirectory(HelloPangolinOffscreen)
     add_subdirectory(HelloPangolinThreads)
     add_subdirectory(SimpleMultiDisplay)
     add_subdirectory(SimpleDisplayImage)
--- a/examples/HelloPangolinOffscreen/CMakeLists.txt
+++ b/examples/HelloPangolinOffscreen/CMakeLists.txt
@@ -1,0 +1,6 @@
+# Find Pangolin (https://github.com/stevenlovegrove/Pangolin)
+find_package(Pangolin 0.5 REQUIRED)
+include_directories(${Pangolin_INCLUDE_DIRS})
+
+add_executable(HelloPangolinOffscreen main.cpp)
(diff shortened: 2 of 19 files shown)


christian-rauch:
The build fails because EGL is missing on Windows and Mac OS. Any ideas how to install EGL on these platforms? Or should offscreen rendering stay a Linux-only feature?

christian-rauch:
The headless EGL context only builds now if EGL is available.

MaverickPeter:
I encountered a problem running the HelloPangolinOffscreen example, it showed that "`what():  No known window handler for URI 'headless'`" and causes core dumped. I wonder if you can help me with this problem.

christian-rauch:
@MaverickPeter What operation system and version are you using? The headless mode is only built if EGL is available. Can you check the output of `OpenGL_EGL_FOUND`, e.g. by `message("EGL available? " ${OpenGL_EGL_FOUND})`?

MaverickPeter:
@christian-rauch  Thanks for your reply. I checked and found that the EGL is not available, but I don't know how to install EGL without a superuser permission, sigh~.

stevenlovegrove:
This is fantastic, thanks so much! The one change I would like to make is to put the window scheme in the existing Params() struct - do you think that is reasonable?

stevenlovegrove:
hmm - that's a pretty good idea I think. It feels like it should be backwards compatible.

christian-rauch:
I moved the `scheme` attribute of the `Uri` class to the parameter map of `Params`.

PR #465, opened 2019-01-03
Fixing an issue with windows top bar

Erick-Mendez:
On Windows this fixes an issue in which the UI elements are offset by the size of the top Windows bar. Getting the actual client dimensions seems to fix it. See images for comparison before and after.

before

after


stevenlovegrove:
This is great, thank you! I will try to verify and merge soon. @scramsby - have you noticed this?

scramsby:
@stevenlovegrove I have noticed this problem recently. I verified locally that this change fixes the issue so I'd recommend to accept the PR.

PR #471, opened 2019-01-15
Allow to override CXX_STANDARD from command line or parent project.

### LINKED ISSUE #469: Setting C++ standard version from outside
I build Pangolin as an external dependency and would like to the standard version to 17 without modifying the source. Currently, it seems that the line

```cmake
set (CMAKE_CXX_STANDARD 14)
```

always overwrites what I set from the command line. If I change it to something like

```cmake
if(NOT CMAKE_CXX_STANDARD)
    set (CMAKE_CXX_STANDARD 14)
endif()
```

it seems to do the trick but still default to `14` if `CMAKE_CXX_STANDARD` is not set.

Can this change be included (I can send a PR)? Or is the another / a better way to set the C++ standard version from outside?

NikolausDemmel:
Fixes #469.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,7 +21,9 @@ option( BUILD_TESTS "Build Tests" ON )
 option( BUILD_TOOLS "Build Examples" ON )
 option( BUILD_EXAMPLES "Build Tools" ON )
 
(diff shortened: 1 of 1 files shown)


NikolausDemmel:
Any chance this change may be considered for merge?

PR #477, opened 2019-02-06
Fix Eigen issue in ModelViewer

martinruenz:
I noticed that the ModelViewer tool was crashing for some valid OBJ files and only in a release build, with the following message: 

> terminate called after throwing an instance of 'std::invalid_argument'
  what():  'Look' and 'up' vectors cannot be parallel when calling ModelViewLookAt.

I could trace this down to a _nan_ in the **view** vector here https://github.com/stevenlovegrove/Pangolin/blob/8cd484a782f4f37a7d5ec61cfcbf453ee5f7a12f/tools/ModelViewer/main.cpp#L119 (even though having solely sensible values when calculating the **view** vector).

It turns out that using the _auto_ keyword in combination with Eigen is known to provoke issues, as described in further detail here: https://eigen.tuxfamily.org/dox/TopicPitfalls.html (see: _C++11 and the auto keyword_)

Replacing _auto_ with _Eigen::Vector3f_ fixed the issue. (Btw I tested with Eigen 3.3.7 and 3.3.4)

### DIFF
--- a/tools/ModelViewer/main.cpp
+++ b/tools/ModelViewer/main.cpp
@@ -115,8 +115,8 @@ int main( int argc, char** argv )
                 auto geom = future_geom.get();
                 auto aabb = pangolin::GetAxisAlignedBox(geom);
                 total_aabb.extend(aabb);
-                const auto center = total_aabb.center();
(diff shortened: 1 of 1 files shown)


stevenlovegrove:
Great, thanks! I was wondering what the cause of that was.

PR #484, opened 2019-02-23
offscreen rendering to FBO

christian-rauch:
Some OpenGL drivers (e.g. intel) seem to have issues with rendering into a pixel buffer (`EGL_SURFACE_TYPE` set to `EGL_PBUFFER_BIT`).
This PR extends the offscreen rendering example with rendering directly into an FBO without relying on the pixel buffer.


stevenlovegrove:
Thanks!

PR #490, opened 2019-03-20
Replace MAC_OS_X_VERSION_10_12 by its value 101200

### LINKED ISSUE #327: mac os error "use of undeclared identifier 'NSWindowStyleMaskClosable'"
when I build it on mac os ,error occurred 
```
error: 
      use of undeclared identifier 'NSWindowStyleMaskClosable'
  ...NSClosableWindowMask backing:NSBackingStoreBuffered defer:YES];
     ^
/Users/lgy/Documents/workspace/Pangolin/src/display/device/display_osx.mm:41:39: note: 
      expanded from macro 'NSClosableWindowMask'
#  define NSClosableWindowMask        NSWindowStyleMaskClosable
                                      ^
5 errors generated.
make[2]: *** [src/CMakeFiles/pangolin.dir/display/device/display_osx.mm.o] Error 1
make[1]: *** [src/CMakeFiles/pangolin.dir/all] Error 2
make: *** [all] Error 2
```

NikolausDemmel:
This check was ineffective on macos versions prior to 10.12, since the
constant is not defined there, which caused build failures due to
undefined symbols. By replacing it with the integral value, the
defines should correctly be skipped on macos < 10.12, while behaviour
should remain unchanged on newer versions.

Fixes #327

### DIFF
--- a/src/display/device/PangolinNSApplication.mm
+++ b/src/display/device/PangolinNSApplication.mm
@@ -29,7 +29,7 @@
 #include <pangolin/display/display.h>
 #include <pangolin/display/device/PangolinNSApplication.h>
 
-#if MAC_OS_X_VERSION_MAX_ALLOWED >= MAC_OS_X_VERSION_10_12
+#if MAC_OS_X_VERSION_MAX_ALLOWED >= 101200
 #  define NSAnyEventMask NSEventMaskAny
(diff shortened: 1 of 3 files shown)


stevenlovegrove:
Thanks!

PR #516, opened 2019-06-10
Unify `CreateWindowAndBind` and `CreateWindowAndBindWithParams`

martinruenz:
I don't see a good reason to split the C++ function `CreateWindowAndBind` into two functions for the python binding (`CreateWindowAndBind` and `CreateWindowAndBindWithParams`), since it is possible to use the same default parameter for the argument `params` in python. To this end, the binding of `Params` simply has to be provided before the binding of `Display`.

This commit removes the python binding `CreateWindowAndBindWithParams` by providing the same functionality with `CreateWindowAndBind` (as in C++).

Further, the python binding of `Params` supports string types now.

Here is an example of how the two changes can be used:

```python3
import pypangolin as pangolin
params = pangolin.Params()
params.Set("scheme", "headless")
pangolin.CreateWindowAndBind("window", 500, 500, params)
```

### DIFF
--- a/src/python/pypangolin/display.cpp
+++ b/src/python/pypangolin/display.cpp
@@ -34,17 +34,12 @@ namespace py_pangolin {
   
   void bind_display(pybind11::module &m) {
     m.def("CreateWindowAndBind",
-          [](const std::string& title, int w=640, int h=480) -> pangolin::WindowInterface& {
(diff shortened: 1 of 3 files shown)


stevenlovegrove:
Thanks Martin!

PR #519, opened 2019-06-20
Fixing a couple of compatibility issues

antimax:
1. Fixing an inconsistent usage of 'new []' and 'delete' operators
in utils/threadedfilebuf.cpp.

This inconsistency results in undefined behavior and, potentially,
memory leaks.

2. Removing the explicit 'mpark' namespace specification in
geometry/geometry_ply.cpp.

The `pangolin/geometry/geometry_ply.h -> pangolin/compat/variant.h`
header sequence is supposed to map `get_if` to eiter `std::get_if`
or `mpark::get_if`, depending on the compiler version. This explicit
namespace specification breaks this compatibility behavior.


antimax:
Apparently, the gcc C++ build for Mac failed in Travis CI because homebrew displayed a donation banner.

PR #520, opened 2019-06-22
Wayland xdg_shell

christian-rauch:
I updated the windowing class for Wayland to support the new `xdg_shell` in place of the deprecated `wl_shell`.

Additional changes:
- add Windows to Travis CI
- update third-party dependencies (pybind11, stb_truetype)
- update the documentation
- minor fixes for compiler warnings

Let me know if you want to have those things in different PRs. E.g. if you want to squash the changes, it would make sense to separate the Wayland stuff from the minor fixes.


stevenlovegrove:
This is great, thank you!

src/video/drivers/openni.cpp
```diff
@@ -80,6 +80,7 @@ OpenNiVideo::OpenNiVideo(OpenNiSensorType s1, OpenNiSensorType s2, ImageDim dim,
         switch( sensor_type[i] ) {
         case OpenNiDepth_1mm_Registered:
             depth_to_color = true;
+            break;
```
stevenlovegrove: I'm doing some code merging and noticed this change again - are you sure it's right? I think the fall through from before was intentional, since use_depth should be true for this case. Was this a fix you tested? I don't have an OpenNI device to test it with.

christian-rauch: No. I might have overeagerly commit this to fix a warning. While working on the main PR.
I don't use the OpenNI interface in Pangolin directly. I actually did not test this change.

I am sorry for introducing this issue. I opened a PR to revert this commit: https://github.com/stevenlovegrove/Pangolin/pull/586

PR #523, opened 2019-07-05
add wayland-protocols to CI

christian-rauch:
Small fix to add `wayland-protocols` to CI and install apt packages without suggestions and recommendations.


christian-rauch:
@stevenlovegrove Can you have a quick look at this? It shouldn't have a big impact on the rest of the code base but makes sure that the Wayland backend is tested.

PR #527, opened 2019-07-11
Minor change - Fixing typo in pybind module

fulkast:
The rotation about Z-axis was incorrectly calling a rotation about the x-axis


stevenlovegrove:
Oops, thanks!

PR #530, opened 2019-08-06
fix Plotter::SetView; not updating target means it zooms back

NikolausDemmel:
Without this, SetView is broken. Since target is not updated, the zooming mechanism will go back to the previous view (or current target view) right after calling SetView.


NikolausDemmel:
The CI failure seems unrelated to the changes. Maybe the job should be restarted.

PR #535, opened 2019-08-12
Jpeg error handler

### LINKED ISSUE #534: jpeg error handling
When an invalid JPEG file is loaded, Pangolin exits with an error message like `Not a JPEG file: starts with 0xff 0xd9`. There is no way to catch this error from a user of Pangolin.
I assume that error is generated within `libjpeg-turbo`. Pangolin should register an error handler and throw an exception that can be caught by user code.

christian-rauch:
Fixes #534 by forwarding the JPEG error to an exception that can be handled by user code. Also includes a fix to print useful OpenGL error messages.

### DIFF
--- a/include/pangolin/gl/glinclude.h
+++ b/include/pangolin/gl/glinclude.h
@@ -37,9 +37,9 @@
 namespace pangolin {
 inline void _CheckGlDieOnError( const char *sFile, const int nLine )
 {
-    GLenum glError = glGetError();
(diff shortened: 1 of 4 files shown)


christian-rauch:
@stevenlovegrove Can you have a look?

stevenlovegrove:
Sorry for the lead time. This is super helpful, thanks so much!

PR #545, opened 2019-09-13
set policy CMP0072 to prefer GLVND for OpenGL_GL_PREFERENCE

christian-rauch:
Sets CMake policy CMP0072 to silence warning about unspecified `OpenGL_GL_PREFERENCE` for newer cmake > 3.10.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -138,6 +138,10 @@ if(BUILD_PANGOLIN_GUI AND BUILD_PANGOLIN_VARS AND BUILD_PANGOLIN_VIDEO )
     list(APPEND SOURCES tools/video_viewer.cpp)
 endif()
 
+if(LINUX)
+    set(OpenGL_GL_PREFERENCE "GLVND")
+endif()
+
 #######################################################
 ## Setup required includes / link info
 


stevenlovegrove:
I don't quite understand the implications of this one - what are the alternatives?

christian-rauch:
The alternative to setting the policy is to set `OpenGL_GL_PREFERENCE` directly by:
```cmake
set(OpenGL_GL_PREFERENCE "GLVND")
```

FindOpenGL on newer CMake versions >3.10 uses GLVND on Linux:
> Some Linux systems utilize GLVND as a new ABI for OpenGL. GLVND separates context libraries from OpenGL itself; OpenGL lives in “libOpenGL”, and contexts are defined in “libGLX” or “libEGL”. GLVND is currently the only way to get OpenGL 3+ functionality via EGL in a manner portable across vendors. Projects may use GLVND explicitly with target OpenGL::OpenGL and either OpenGL::GLX or OpenGL::EGL.

(source: https://cmake.org/cmake/help/latest/module/FindOpenGL.html#linux-specific)

If the policy or `OpenGL_GL_PREFERENCE` is not set, it will show a warning:
```
  Policy CMP0072 is not set: FindOpenGL prefers GLVND by default when
  available.  Run "cmake --help-policy CMP0072" for policy details.  Use the
  cmake_policy command to set the policy and suppress this warning.

  FindOpenGL found both a legacy GL library:

    OPENGL_gl_LIBRARY: /usr/lib/x86_64-linux-gnu/libGL.so

  and GLVND libraries for OpenGL and GLX:

    OPENGL_opengl_LIBRARY: /usr/lib/x86_64-linux-gnu/libOpenGL.so
    OPENGL_glx_LIBRARY: /usr/lib/x86_64-linux-gnu/libGLX.so

  OpenGL_GL_PREFERENCE has not been set to "GLVND" or "LEGACY", so for
  compatibility with CMake 3.10 and below the legacy GL library will be used.
```

stevenlovegrove:
Perhaps we can use if(LINUX) set(OpenGL_GL_PREFERENCE "GLVND") instead? Can you think of any negatives?

I hate these policy commands because they are super opaque!

christian-rauch:
I agree. I am usually also not a fan of setting these anonymously named policies.

stevenlovegrove:
Thanks!

PR #550, opened 2019-09-30
Scroll in LayoutVertical with not shown displays

caselitz:
Sorry for creating a new pull request, it's identical with [this one](https://github.com/stevenlovegrove/Pangolin/pull/429).

### DIFF
--- a/include/pangolin/display/view.h
+++ b/include/pangolin/display/view.h
@@ -64,7 +64,7 @@ struct PANGOLIN_EXPORT View
 {
     View(double aspect=0.0)
         : aspect(aspect), top(1.0),left(0.0),right(1.0),bottom(0.0), hlock(LockCenter),vlock(LockCenter),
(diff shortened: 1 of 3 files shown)


stevenlovegrove:
I know I haven't set a good example with the private variables, but could you make scroll_show private? I understand now what you're doing, and it is kinda a hack to work around my poor hack - I just want to be careful this isn't used as part of the View interface by anyone externally, so that it can be refactored away eventually.

caselitz:
Thanks for the reply Steven. I tried to follow the code that is there, but I agree that it makes sense to make `scroll_show` private since public use is not intended. Unfortunately `scroll_offset` is used in [`HandlerScroll`](https://github.com/stevenlovegrove/Pangolin/blob/master/src/handler/handler.cpp#L92-L115) so it needs to stay public.

I replaced the PR commit with a new one that has `scroll_show` private.

stevenlovegrove:
Thanks for your perseverance to merge this :)

PR #558, opened 2019-10-09
Add further opengl python bindings

martinruenz:
This pull request exposes some additional functions of `GlFramebuffer` and `GlTexture` to python.


martinruenz:
Yay! :tada: :smile:
I still rely on those bindings from time to time, so that's really useful for me. Thanks!

PR #564, opened 2019-10-28
Feature/colour provider

Erick-Mendez:
Adds an abstraction for color providers used by the plotter. ColourWheel becomes a provider, and it also adds a simple ColourCircularBuffer provider. The plotter constructor is overloaded to provide backwards compatibility, with the default being a ColourWheel provider (as before).

Same as before with the ColourWheel:
image

Using the simple ColourCircularBuffer (and shades of green):
image

The SimplePlot example is updated to showcase how this functionality works:

```
  bool use_wheel = true; /// By default we use a ColourWheel provider

  if (use_wheel) {
      colours = std::make_unique<pangolin::ColourWheel>(0.6);
  }
  else {
      colours = std::make_unique<pangolin::ColourCircularBuffer>();
      colours->Add(pangolin::Colour::Green().WithAlpha(0.3f));
      colours->Add(pangolin::Colour::Green().WithAlpha(0.6f));
      colours->Add(pangolin::Colour::Green().WithAlpha(0.9f));
  }

  // OpenGL 'view' of data. We might have many views of the same data.
  pangolin::Plotter plotter(&log, std::move(colours), 0.0f,4.0f*(float)M_PI/tinc,-2.0f,2.0f,(float)M_PI/(4.0f*tinc),0.5f);

```

### DIFF
--- a/components/pango_opengl/include/pangolin/gl/colour.h
+++ b/components/pango_opengl/include/pangolin/gl/colour.h
@@ -135,11 +135,26 @@ struct Colour
 
 };
 
+/// The base class of colour providers. It does not enforce any strategy on how to provide the colours
+class ColourProvider
+{
+public:
+    /// Adds a colour to this provider. Some providers that generate colours
+    /// might not need to implement this (e.g. ColourWheel)
+    virtual void Add(const Colour& colour) {};
+
+    /// Get next colour. In the case of generation it would create a new one.
+    virtual Colour GetNext() = 0;
+
+    /// Reset colours
+    virtual void Reset() = 0;
+};
+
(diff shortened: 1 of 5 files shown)


Erick-Mendez:
@stevenlovegrove I wonder if you could have a look at this PR. It's a relatively small feature.
(unfortunately, I cannot add reviewers myself).
Thanks!

stevenlovegrove:
Hi Erick - sorry I've never replied on this PR. I really like the idea and have actually been meaning to modify it slightly. My main objection to the current implementation is that it breaks the API. I have also been wanting to support colormaps and have been wondering if there is a unified API that could be used for the discrete and continuous use cases.

PR #565, opened 2019-10-30
cmake: Fix wayland option

stotko:
Support for Wayland has been added in #520. However, the new option `DISPLAY_WAYLAND` does not have any effect if the Wayland client has been found. This leads to generation errors when `xkbcommon` is not installed and `DISPLAY_WAYLAND` is set to `OFF` (which is also the default value). Fix the condition to make the option work as intended.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -212,57 +212,59 @@ else()
             list(APPEND LINK_LIBS "-framework Cocoa" )
         elseif(_LINUX_)
             # Wayland
-            find_package(Wayland QUIET)
-            if(DISPLAY_WAYLAND OR WAYLAND_CLIENT_FOUND)
-                find_package(Wayland REQUIRED)
-                find_package(PkgConfig)
-                pkg_check_modules(xkbcommon REQUIRED xkbcommon)
-                pkg_check_modules(egl QUIET egl)
+            if(DISPLAY_WAYLAND)
+                find_package(Wayland QUIET)
+                if(WAYLAND_CLIENT_FOUND)
+                    find_package(Wayland REQUIRED)
+                    find_package(PkgConfig)
+                    pkg_check_modules(xkbcommon REQUIRED xkbcommon)
+                    pkg_check_modules(egl QUIET egl)
 
-                # find Wayland protocols
-                pkg_get_variable(WAYLAND_PROTOCOLS_DIR wayland-protocols pkgdatadir)
+                    # find Wayland protocols
+                    pkg_get_variable(WAYLAND_PROTOCOLS_DIR wayland-protocols pkgdatadir)
 
-                # generate header for protocol 'xdg_shell'
-                set(XDG_PROT_DEF "${WAYLAND_PROTOCOLS_DIR}/stable/xdg-shell/xdg-shell.xml")
-                if(EXISTS ${XDG_PROT_DEF})
-                    # use the 'xdg_shell' protocol
-                    add_definitions(-DUSE_WL_XDG=1)
+                    # generate header for protocol 'xdg_shell'
+                    set(XDG_PROT_DEF "${WAYLAND_PROTOCOLS_DIR}/stable/xdg-shell/xdg-shell.xml")
+                    if(EXISTS ${XDG_PROT_DEF})
+                        # use the 'xdg_shell' protocol
+                        add_definitions(-DUSE_WL_XDG=1)
 
-                    # find 'wayland-scanner' executable
-                    pkg_get_variable(WAYLAND_SCANNER wayland-scanner wayland_scanner)
+                        # find 'wayland-scanner' executable
+                        pkg_get_variable(WAYLAND_SCANNER wayland-scanner wayland_scanner)
 
-                    # generate protocol implementation
-                    add_custom_command(
-                        OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/xdg-shell-client-protocol.h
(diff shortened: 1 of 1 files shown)


stevenlovegrove:
@christian-rauch - this looks good, right?

src/CMakeLists.txt
```diff
@@ -213,7 +213,7 @@ else()
         elseif(_LINUX_)
             # Wayland
             find_package(Wayland QUIET)
-            if(DISPLAY_WAYLAND OR WAYLAND_CLIENT_FOUND)
+            if(DISPLAY_WAYLAND AND WAYLAND_CLIENT_FOUND)
```
christian-rauch, line 216: The idea with the "OR" was that the Wayland support is enabled if either the user activates this manually by setting `DISPLAY_WAYLAND` or automatically if the Wayland development libraries are installed (`WAYLAND_CLIENT_FOUND`). If this has been detected, the CMake errors give a hint which remaining libraries have to be installed.

I would propose to either fix the automatic behaviour by adding the remaining libs to the condition, e.g. `if(DISPLAY_WAYLAND OR (WAYLAND_CLIENT_FOUND AND WAYLAND_EGL_FOUND AND WAYLAND_CURSOR_FOUND AND ...))` or by removing them entirely and just rely on the manual activation, i.e. by just `if(DISPLAY_WAYLAND)`.

I think the "AND" connection does not make sense since you probably want to get notified when libraries are missing but you requested Wayland support manually by setting `DISPLAY_WAYLAND`.

stotko: The intention for `AND` was to mimic the check for `X11` below. I changed it to follow your second suggestion which is in line with the behavior for the other optional dependencies (Python, Eigen, etc.). 

The diff is a bit larger now because of the additional `if` condition and the resulting indentation. Please take a look again.

christian-rauch: I actually thought that if we enable Wayland only by user choice (`DISPLAY_WAYLAND`) then we can just require all the dependencies (find_package(... REQUIRED)) and don't have to be conditional on them. I.e. something like `find_package(Wayland QUIET)` and `if(WAYLAND_CLIENT_FOUND)` can be removed and CMake should just fail when a Wayland dependency hasn't been found.

stotko: I agree that this is a bit redundant. However, like all other dependencies (except OpenGL), Wayland is documented as an **optional dependency** and should follow the same behavior as the others. That is, if the user wants to have that features (`PANGOLIN_BUILD_<LIB>` is `ON`) and the library has been found (`<LIB>_FOUND`), then enable it. Otherwise, disable it and do not throw any errors:
```
if(PANGOLIN_BUILD_<LIB>)
  find_package(<LIB> QUIET)
  if(<LIB>_FOUND)
     # Enable <LIB> by adding its include directory, link libraries
     # as well as the related Pangolin source files
  endif()
endif()
```

I am not sure what components are required for Wayland support, but the logic here looks more complicated than for the other dependencies as it searches for additional libraries after Wayland-Client has been found. Since this looks intentional, I just kept it and changed the build option to follow the pattern mentioned above.

christian-rauch: You are right about having this optional, but since it's deactivated by default these checks wouldn't be done anyway until someone actively chooses to activate that support. It would be useful to get error messages once you chose a feature actively, but its dependencies cannot be found. But I now see that this is indeed the same behaviour as for the X11 UI, which also silently is deactivated if not found.

The Wayland support requires more than one package to be installed: `sudo apt install libegl1-mesa-dev libwayland-dev libxkbcommon-dev wayland-protocols`.
Currently, the support would be activated in CMake if `libwayland-dev` is installed and it would then check for the remaining dependencies. So, if you want to replicate the X11 case in the CMake file, you have to check for all of them.

christian-rauch:
I am fine with the changes now since the error will only be shown if `DISPLAY_WAYLAND` has been selected and `libwayland-dev` is installed. I assume a user interested in the Wayland support can deal with the issue of installing the remaining dependencies.

src/CMakeLists.txt
stotko: BTW, `FindWayland.cmake` also complains about non-standard usage of `find_package_handle_standard_args`. It is usually intended to be exclusively used with the package name i.e. `Wayland` in this case, see [its documentation](https://cmake.org/cmake/help/latest/module/FindPackageHandleStandardArgs.html). In contrast, it is also used to detect the [individual components](https://github.com/stevenlovegrove/Pangolin/blob/master/CMakeModules/FindWayland.cmake#L52) which results in many thrown warnings.

Considering the side note above, I would suggest the following:
1. Apply this fix in its current form to match the bahavior of the other optional dependencies.
2. In a follow-up PR, move the `find_X` commands to find the additionally required dependencies, i.e. `xkbcommon`, etc. into `FindWayland.cmake` and use `find_package_handle_standard_args` to detect if **all** dependencies have been found.

This way, `find_package(Wayland QUIET)` would actually mean "all Wayland-specific dependencies required for Pangolin" rather than "only the basic Wayland libraries".

christian-rauch: The `FindWayland.cmake` is taken from another project and only checks for the Wayland libraries. For its usage in Pangolin it would indeed make sense to extend the set of required packages.

stotko: This cleanup for `FindWayland.cmake` would also allow to set the default for `DISPLAY_WAYLAND` to `ON` by default since it now follows the same convention as all other modules.

stevenlovegrove:
Thanks for drilling down on this, folks. Merging.

PR #566, opened 2019-10-30
cmake: Fix Linux platform check

stotko:
Recently, a new CMake warning concerning OpenGL has been fixed (see #545). However, the variable `LINUX` is not defined which makes the warning still appear. Use `_LINUX_` instead which is defined in `CMakeModules/SetPlatformVars.cmake` to fix this issue.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -124,7 +124,7 @@ if(BUILD_PANGOLIN_VIDEO)
     RegisterThreadVideoFactory
   )
 
-  if(LINUX)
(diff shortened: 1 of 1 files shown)


stotko:
Can this PR be merged as well to make the fix in #545 work as intended? Without it, the annoying CMake policy [CMP0072](https://cmake.org/cmake/help/latest/policy/CMP0072.html#policy:CMP0072) warning is still thrown.

stevenlovegrove:
Thanks!

PR #569, opened 2019-11-29
set Python 3 shebang for example scripts

christian-rauch:
shebangs for conveniently running scripts as executables and without guessing the required Python version :-)


christian-rauch:
Ping @stevenlovegrove . This is a trivial change the simplifies running the python examples.

PR #571, opened 2019-12-06
geometry: Add alternative PLY property types

stotko:
The PLY loader defines the set of supported PLY property types. Although all possible scalar types [1] are defined and supported (char, uchar, ..., double), there is also an alternative/alias type [2,3] for each of them (int8,uint8, ..., float64) that is not included here. Since many other libraries and tools support both formats, a user might be confused why certain files that silently use one of these alias types will be rejected by Pangolin. Extend the list of property types to fix this issue.

[1] http://paulbourke.net/dataformats/ply/
[2] http://people.math.sc.edu/Burkardt/data/ply/ply.txt
[3] https://en.wikipedia.org/wiki/PLY_(file_format)


stevenlovegrove:
Perfect, thanks! (Sorry for the glacially slow review)

PR #576, opened 2020-02-04
deactivate OpenNI by default

christian-rauch:
Reduce compile dependencies.
Can be activated again by `-DBUILD_PANGOLIN_OPENNI=ON -DBUILD_PANGOLIN_OPENNI2=ON`.


stevenlovegrove:
The Pangolin ethos is generally "enable it if it's there". If you disable this by default, why not everything else?

christian-rauch:
There are many hardware-specific dependencies in Pangolin (OpenNI, RealSense, DepthSense, TeliCam, ...) which are activated by default. It is very unlikely that a user will need all of them at once inside Pangolin. It would be more efficient to just have those enabled that a user actually uses.

This, of course, doesn't mean that these libraries can't be used manually to stream images to Pangolin.

If you still prefer to have these hardware dependencies enables by default, I will just close this PR.

stevenlovegrove:
My view is that if people have these SDK's on their computer, it is likely that they will wish to use them with Pangolin. A compromise might be to introduce a 'NO_DEFAULT_VIDEO_DRIVERS' flag or similar to override this behaviour.

christian-rauch:
Fair point. I am going to close this and manually deactivate the flags if I require a faster compilation.

PR #598, opened 2020-05-13
fix compiler warning on __STDC_FORMAT_MACROS redef

NikolausDemmel:
Without this, if I include 2 libraries that define this macro, I get warnings like

```
warning: '__STDC_FORMAT_MACROS' macro redefined [-Wmacro-redefined]
```

This is on macos 10.15 with AppleClang 11.

See also upstream PR: https://github.com/kazuho/picojson/pull/127


stevenlovegrove:
Thanks!

PR #600, opened 2020-05-21
Add python bindings for DrawPoints()

markoelez:
Added python bindings for the `DrawPoints()` method.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,25 @@ can execute:
 cmake --build . --target pangolin_doc
 ```
 
+## Building Python bindings ##
+
+```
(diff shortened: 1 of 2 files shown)


stevenlovegrove:
Thanks for this, but I would rather wrap the C++ methods with pybind11 in order to reduce implementation code and increase performance.

A glDrawPoints binding did manage to make it in here ab14cf9212b9be659427ada4d8e9f54d9f48b316 in case you still need this.

chaitanyantr:
can I give same params for glDrawpoints as Drawpoints? when I replace drawpoints with gldrawpoints it gives error.
errr

Issue #623, opened 2020-10-08
flipped screencast


christian-rauch:
The `RecordOnRender` records videos upside-down.

This can easily be reproduced with the `SimpleDisplay` example with the "Record_Cube" button:
![vlcsnap-2020-10-08-18h05m36s309](https://user-images.githubusercontent.com/8226248/95490968-e5c89300-0990-11eb-918d-01f8e465c23d.png)

christian-rauch:
Well, I guess this is just a matter of adding `flip=true` to the parameters, e.g.;
`pangolin::DisplayBase().RecordOnRender("ffmpeg:[fps=50,bps=8388608,unique_filename,flip=true]//screencap.avi");`.

Is there a reason why the video has to be flipped?

christian-rauch:
According to https://github.com/stevenlovegrove/Pangolin/pull/624#issuecomment-906853607, this is obsolete now.

PR #624, opened 2020-10-12
flip recording

christian-rauch:
A quick fix for the `SimpleDisplay` to record flipped and in mp4 format. This gives an example how to work-around https://github.com/stevenlovegrove/Pangolin/issues/623.


christian-rauch:
@stevenlovegrove The CI cannot access the Eigen archive for Windows anymore.

PR #628, opened 2020-10-29
fix: Correct hit buffer processing in scenehandler

hobbeshunter:
Unnamed elements should be ignored when looking for the closest hit.

Another option would be to replace ProcessHitBuffer with just

```
        for (int hit = 0; hit < hits; hit++) {
            for(unsigned int name = 0; name < buf[0]; name++) {
                const int pickId = (buf + 3)[name];
                hit_map[pickId] = InteractiveIndex::I().Find(pickId);
            }
            buf += buf[0] + 3;
        }
```

That way every element gets the chance to react to the mouse event. But then it would be better to order them by z distance and stop the Mouse callback calling as soon as one indicated handling (returned true).


stevenlovegrove:
This is great, thank you! I'll try to merge this in when I get a chance.

hobbeshunter:
Nice. Glad I could help

PR #658, opened 2021-05-24
Fix GCC 11 compatibility issue

### LINKED ISSUE #657: Naming conflict with GCC 11
I'm trying to compile a project which uses Pangolin, I got an error when compiling with GCC-11:

```
/usr/include/pangolin/utils/picojson.h:273:65: error: macro "_serialize" passed 2 arguments, but takes just 0
  273 |     template <typename Iter> void _serialize(Iter os, int indent) const;
...
...
/usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/include/serializeintrin.h:37: note: macro "_serialize" defined here
   37 | #define _serialize()    __builtin_ia32_serialize ()
      | 
```

In fact, GCC-11 defines a [new macro `_serialize`](https://gcc.gnu.org/gcc-11/changes.html), creating a conflict with the private member defined in `pangolin/utils/picojson.h`.

In general, I think it will be safer to name private members as `something_` rather than `_something` since **all names starting with an underscore are considered as [reserved names in GCC](https://www.gnu.org/software/libc/manual/html_node/Reserved-Names.html)**

abougouffa:
Fixes #657

### DIFF
--- a/include/pangolin/utils/picojson.h
+++ b/include/pangolin/utils/picojson.h
@@ -270,8 +270,8 @@ public:
 private:
     template <typename T> value(const T*); // intentionally defined to block implicit conversion of pointer to bool
     template <typename Iter> static void _indent(Iter os, int indent);
-    template <typename Iter> void _serialize(Iter os, int indent) const;
-    std::string _serialize(int indent) const;
+    template <typename Iter> void serialize_(Iter os, int indent) const;
+    std::string serialize_(int indent) const;
 };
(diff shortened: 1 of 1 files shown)


abougouffa:
Hello @stevenlovegrove 

Can you merge this little PR, I getting this problem each time I rebuild pangolin.

Thanks for the good work

IcneDy (approved):

PR #666, opened 2021-08-27
fix: numeric_limits compilation error

abougouffa:
A small fix for a compilation error with GCC 11.1.0

`numeric_limits` is used in `components/pango_core/src/factory/factory_help.cpp` but the `<limits>` header is not included.


stevenlovegrove:
Thank you! This was included in upstream here: 1658218dfbe56a3994e081324802bb354b12742e

PR #695, opened 2021-10-18
re-enable Wayland

christian-rauch:
This re-enables the Wayland `WindowInterface` with some additional fixes. As before, this is automatically enabled at compile time when the wayland development libraries are found. At runtime, it will first check if a wayland display server Is running and if not (or another error occurs during initialisation) it will fall back to X11.

I also added a new CMake variable `BUILD_ASAN` that enables the [AddressSanitizer](https://clang.llvm.org/docs/AddressSanitizer.html) to check for memory issues in Debug builds.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,6 +23,7 @@ include(PangolinFactory)
 
 option( BUILD_TOOLS "Build Tools" ON )
 option( BUILD_EXAMPLES "Build Examples" ON )
+option( BUILD_ASAN "enable AddressSanitizer for Debug builds to detect memory issues" OFF)
 
 # Default build type (Override with cmake .. -DCMAKE_BUILD_TYPE=...)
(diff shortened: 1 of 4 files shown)


stevenlovegrove:
I've been having trouble with vcpkg in that the versioned packages can break since they can refer to zips and things outside of the repo which is probably also what's happened here.

Is this PR ready to go otherwise, or do you plan to squash?

stevenlovegrove:
Okay, I finally got around to test a windows fix for the vcpkg stuff and merged this PR, thanks! 

I skipped the ASAN addition for now - I would rather do something like https://stackoverflow.com/a/64294837 I think within a cmake utility file.

christian-rauch:
@stevenlovegrove Can you delete the [`wayland_rebase` branch](https://github.com/stevenlovegrove/Pangolin/tree/wayland_rebase) from your upstream repo? I think this is the one you used to merge this PR into master. It's redundant now.

PR #710, opened 2021-11-22
set only one default win URI handler and prefer Wayland over X11

### LINKED ISSUE #709: "PANGO_DEFAULT_WIN_URI" redefined
There is a compiler error about `PANGO_DEFAULT_WIN_URI`:
```
<command-line>: error: "PANGO_DEFAULT_WIN_URI" redefined [-Werror]
```
that appears multiple times on stdout and prevents Pangolin from compiling.

However, on another system, this just appears as a warning:
```
<command-line>: warning: "PANGO_DEFAULT_WIN_URI" redefined
```

Edit: This is likely from another project defining `-Werror`. So a work-around is to remove this, but it would be better to fix the warnings so that other projects do not encounter this.

christian-rauch:
- set only a single `PANGO_DEFAULT_WIN_URI`
- set to `x11` only of not build with Wayland
- set to `wayland` if built with Wayland support, so that Wayland will be preferred automatically and fall back to X11
- Fixes #709

### DIFF
--- a/components/pango_windowing/CMakeLists.txt
+++ b/components/pango_windowing/CMakeLists.txt
@@ -28,9 +28,8 @@ else()
         target_sources( ${COMPONENT} PRIVATE ${CMAKE_CURRENT_LIST_DIR}/src/display_x11.cpp )
         target_link_libraries(${COMPONENT} PRIVATE ${X11_LIBRARIES} )
         target_include_directories(${COMPONENT} PRIVATE ${X11_INCLUDE_DIR} )
+        PangolinRegisterFactory(WindowInterface X11Window)
     endif()
-    target_compile_definitions(${COMPONENT} PUBLIC "PANGO_DEFAULT_WIN_URI=\"x11\"")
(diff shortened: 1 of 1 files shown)


stevenlovegrove:
WAYLAND_CLIENT_FOUND gets set [here](https://github.com/stevenlovegrove/Pangolin/blob/ea85d913dc5d54e38a7ae5bc73b0f8010c6bf298/components/pango_windowing/CMakeLists.txt#L39) presumably
 so the variable you're using isn't available yet.

I'd recommend putting this logic in an else [here](https://github.com/stevenlovegrove/Pangolin/blob/ea85d913dc5d54e38a7ae5bc73b0f8010c6bf298/components/pango_windowing/CMakeLists.txt#L75)

christian-rauch:
> so the variable you're using isn't available yet.

Right. I missed that and CMake did not complain :-)

I moved it to the else block.

stevenlovegrove:
Thanks!

PR #721, opened 2021-12-18
Extend python interface

VladyslavUsenko:
This PR extends the python interface of pangolin:
* more functions from `gldraw`.
* Support for `uint8_t` and `uint16_t` image types in `ImageView.SetView`.
* Expose `GuiChanged` for variables in python. Similar to C++ version it resets the flag on read.
*  Update python example to showcase `GuiChanged` and `Toggle`.


stevenlovegrove:
Thanks!

PR #733, opened 2022-02-11
doc and script update

christian-rauch:
Small updates to the install script and documentation.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -61,17 +61,14 @@ jobs:
         $GITHUB_WORKSPACE/scripts/install_prerequisites.sh -v -u -m ${{matrix.package_manager}} all
 
     - name: Configure CMake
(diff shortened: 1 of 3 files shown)


stevenlovegrove:
I also prefer and use Ninja, but I don't think it's wise to suggest that in the readme as it is another dependency that can go wrong and can confuse people who haven't heard of it

christian-rauch:
Ping @stevenlovegrove .

christian-rauch:
The main reason for this PR is to simplify the command-line steps for building the CMake project. The Ninja steps are now optional.

I also added the simplified build instructions to the CI script. I experimented with getting the Ninja steps running in the CI, but this failed with some compilation error on Windows with OpenNI.

christian-rauch:
@stevenlovegrove Can you have a look at this again? The Ninja build is now an optional step.

PR #739, opened 2022-03-02
implement getters for width and height

edgarriba:
I also noticed that `MjpegVideo` it's not well exposed to the global library namespace.

Besides, the `GetCurrentFrameId()` seems to return 0 all the time.

### DIFF
--- a/components/pango_video/include/pangolin/video/drivers/mjpeg.h
+++ b/components/pango_video/include/pangolin/video/drivers/mjpeg.h
@@ -63,6 +63,9 @@ public:
     size_t GetTotalFrames() const override;
     size_t Seek(size_t frameid) override;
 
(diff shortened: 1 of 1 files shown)


stevenlovegrove:
MpegVideo isn't intended to be used directly. It can be instantiated using the factory and behaves as with the VideoInterface and VideoPlaybackInterface as an opaque implementation.

As with all the VideoInterface devices, you can access image dimensions with the Streams() member function.

If you use the VideoInput wrapper, that does have a Width() and Height() convenience method.

Regarding the frame number, you're right - it looks like I'm missing an increment on the grab method. It should change with each seek though.

edgarriba:
thanks so much! got fixed :)

PR #750, opened 2022-03-23
disable tests in default configuration

christian-rauch:
Disable tests by default in the CMake and enable them in the CI.


stevenlovegrove:
Thanks!

Issue #737, opened 2022-02-22
compilation with ffmpeg 5.0


NikolausDemmel:
Looks like compilation is broken with the recently released ffmpeg 5.0 (see below). Since homebrew on macOS is rolling release, Pangolin doesn't compile on macOS out-of-the-box any more.

If you use cmake >= 3.12 (which on macOS should generally be the case) you can work around the issue by pointing cmake to ffmpeg v4 explicitly:

```
# make sure ffmpeg version 4 is installed
brew install ffmpeg@4

# explicitly point cmake to the right version of ffmpeg
cmake .. -DFFMPEG_ROOT=`brew --prefix`/opt/ffmpeg@4/ -DCMAKE_POLICY_DEFAULT_CMP0074=NEW
```

But it would be nice if Pangolin would work with ffmpeg 5.0.

This is the compilation error with current `master` on macOS Catalina with AppleClang 12, when it builds against the default ffmpeg. Looks like some API with codecs changed (or deprecated API was removed). See this [blog post](http://www.jbkempf.com/blog/post/2022/FFmpeg-5.0) about the release (or [google cached version](http://webcache.googleusercontent.com/search?q=cache:DdST4P4hE38J:www.jbkempf.com/blog/post/2022/FFmpeg-5.0+&cd=1&hl=en&ct=clnk&gl=de), since the blog post appears to be down).
```
cmake ..
[...]
make
[...]
[ 36%] Building CXX object CMakeFiles/pango_video.dir/components/pango_video/src/drivers/ffmpeg.cpp.o
In file included from /Users/demmeln/work/slam/Pangolin/components/pango_video/src/drivers/ffmpeg.cpp:37:
In file included from /Users/demmeln/work/slam/Pangolin/components/pango_video/include/pangolin/video/drivers/ffmpeg.h:32:
/Users/demmeln/work/slam/Pangolin/components/pango_video/include/pangolin/video/drivers/ffmpeg_common.h:97:5: error: use of undeclared identifier 'AV_PIX_FMT_VAAPI_MOCO'; did you mean 'AV_PIX_FMT_VAAPI'?
    TEST_PIX_FMT_RETURN(VAAPI_MOCO);
    ^
/Users/demmeln/work/slam/Pangolin/components/pango_video/include/pangolin/video/drivers/ffmpeg_common.h:37:39: note: expanded from macro 'TEST_PIX_FMT_RETURN'
[...]

PR #754, opened 2022-04-09
cmake: Output dependency location; add BUILD_PANGOLIN_PYHTHON; remove Eigen find module

NikolausDemmel:
Three small self-contained tweaks (I can also do separate PRs if you prefer).

1. Output the header location for the found Eigen and ffmpeg locations. This is very useful when you have Pangolin included in your project and you use a custom Eigen version, to ensure the dependency also finds the correct Eigen. Output for ffmpeg was very useful in the wake of #737. (Other dependencies could get similar output, but I haven't needed those so far.)
2. Bring back the `BUILD_PANGOLIN_PYHTHON` (was removed in the component refactoring) to build w/o Python even if it's found. Makes for a leaner build if you don't need Python.
3. Remove the find module for Eigen and at the same time don't force CONFIG mode when finding Eigen. Currently, the Eigen find module seems to be unused now, b/c find_package(Eigen3...) forces CONFIG. However, the generated config files don't force CONFIG for Eigen and thus they do use the find module (see `cmake/PangolinConfig.cmake.in`). I suggest the most consistent behaviour would be removing the find module and not forcing CONFIG. This also allows projects that include Pangolin to customize the Eigen location by supplying their own find module.


stevenlovegrove:
Sounds like that Eigen fix will be pretty useful, thanks!

PR #764, opened 2022-05-15
Add image rotation option to handler_image

Dvad:
This adds methods to rotate the texture before display in handler_image. Rotation are multiple of 90degrees and can be set by using the `SetThetaQuarterTurn` method.

This changes takes care of making the mouse interaction consistent with the rotation for instance when selecting.
The user should call "glSetModelView" before rendering opengl primitive in the original pixel grid. The function will apply the correction rotation.

Test:
As example, here is a random image draw with the four possible orientations. We draw a white line between pixel (0,0) and (1,1).
With every chosen rotation, the line is correctly displayed.
Screenshot from 2022-05-14 17-13-40
Screenshot from 2022-05-14 17-13-23
Screenshot from 2022-05-14 17-13-09
Screenshot from 2022-05-14 17-12-52

Code snippet for testing:
```c++
#include <limits>
#include <iostream>

#include <pangolin/display/display.h>
#include <pangolin/display/view.h>
#include <pangolin/handler/handler.h>
#include <pangolin/handler/handler_image.h>
#include <pangolin/gl/gldraw.h>

void setImageData(unsigned char * imageArray, int size){
  for(int i = 0 ; i < size;i++) {
    imageArray[i] = (unsigned char)(rand()/(RAND_MAX/255.0));
  }
}

int main(/*int argc, char* argv[]*/)
{
  // Create OpenGL window in single line
  pangolin::CreateWindowAndBind("Main",640,480);

  // 3D Mouse handler requires depth testing to be enabled
  glEnable(GL_DEPTH_TEST);
  
  pangolin::OpenGlRenderState s_cam(
    pangolin::ProjectionMatrix(640,480,420,420,320,240,0.1,1000),
    pangolin::ModelViewLookAt(-1,1,-1, 0,0,0, pangolin::AxisY)
  );

  // Aspect ratio allows us to constrain width and height whilst fitting within specified
  // bounds. A positive aspect ratio makes a view 'shrink to fit' (introducing empty bars),
  // whilst a negative ratio makes the view 'grow to fit' (cropping the view).
  pangolin::View& d_cam = pangolin::Display("cam")
      .SetBounds(0,1.0f,0,1.0f,-640/480.0)
      .SetHandler(new pangolin::Handler3D(s_cam));

  // This view will take up no more than a third of the windows width or height, and it
  // will have a fixed aspect ratio to match the image that it will display. When fitting
  // within the specified bounds, push to the top-left (as specified by SetLock).
  pangolin::View& d_image = pangolin::Display("image")
      .SetBounds(2/3.0f,1.0f,0,1/3.0f,1.f)
      .SetLock(pangolin::LockLeft, pangolin::LockTop);

  std::cout << "Resize the window to experiment with SetBounds, SetLock and SetAspect." << std::endl;
  std::cout << "Notice that the cubes aspect is maintained even though it covers the whole screen." << std::endl;

  const int width =  5;
  const int height = 5;

  unsigned char* imageArray = new unsigned char[3*width*height];
  pangolin::GlTexture imageTexture(width,height,GL_RGB,false,0,GL_RGB,GL_UNSIGNED_BYTE);
  pangolin::ImageViewHandler handler(width, height);
  d_image.SetHandler(&handler);
  // Default hooks for exiting (Esc) and fullscreen (tab).
  setImageData(imageArray,3*width*height);
  int i = 0;
  while(!pangolin::ShouldQuit())
  {
    glDisable(GL_DEPTH_TEST);
    glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
    //Set some random image data and upload to GPU
    imageTexture.Upload(imageArray,GL_RGB,GL_UNSIGNED_BYTE);
    //display the image
    d_image.Activate();
    glColor3f(1.0,1.0,1.0);
    //imageTexture.RenderToViewport();
    handler.SetThetaQuarterTurn(i/60);
    handler.UpdateView();
    handler.glSetViewOrtho();
    handler.UseNN() = true;
    handler.glRenderTexture(imageTexture);
    handler.glRenderOverlay();
    handler.glSetModelView();
    pangolin::glDrawLine(0,0,1,1);
    pangolin::FinishFrame();
    i++;
  }

  delete[] imageArray;

  return 0;
}
```

### DIFF
--- a/components/pango_display/src/handler_image.cpp
+++ b/components/pango_display/src/handler_image.cpp
@@ -5,7 +5,7 @@ namespace pangolin
 {
 
 ImageViewHandler::ImageViewHandler(const std::string & title)
-    : linked_view_handler(0), use_nn(false), flipTextureX(false), flipTextureY(false), title(title)
+    : linked_view_handler(0), use_nn(false), flipTextureX(false), flipTextureY(false), thetaQuarterTurn(0), title(title)
 {
     SetDimensions(1, 1);
 }
@@ -17,12 +17,27 @@ ImageViewHandler::ImageViewHandler(size_t w, size_t h)
     SetDimensions(w,h);
 }
 
+void ImageViewHandler::SetRviewDefaultAndMax()
+{
+  if (GetThetaQuarterTurn()%2==0)
+  {
+    rview_default = pangolin::XYRangef(-0.5f, original_img_width-0.5f, -0.5f, original_img_height-0.5f),
+    rview_max = pangolin::XYRangef(-0.5f, original_img_width-0.5f, -0.5f, original_img_height-0.5f);
+  }
+  else{
+    rview_default = pangolin::XYRangef(-0.5f, original_img_height-0.5f, -0.5f, original_img_width-0.5f),
+    rview_max = pangolin::XYRangef(-0.5f, original_img_height-0.5f, -0.5f, original_img_width-0.5f);
+  }
+  rview = rview_default;
+  target = rview;
+}
+
 void ImageViewHandler::SetDimensions(size_t w, size_t h)
 {
-    rview_default = pangolin::XYRangef(-0.5f, w-0.5f, -0.5f, h-0.5f),
-    rview_max = pangolin::XYRangef(-0.5f, w-0.5f, -0.5f, h-0.5f),
-    rview = rview_default;
-    target = rview;
+    // w and h are dimensions of the input image
+    original_img_width = w;
+    original_img_height = h;
+    SetRviewDefaultAndMax();
 }
 
 void ImageViewHandler::UpdateView()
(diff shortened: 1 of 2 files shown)


christian-rauch:
Can you add the example directly to the repo? It would be nice to indirectly document this by having it in one of the examples.

You could add an `ImageViewHandler` to the `SimpleDisplayImage` example and use the new `SetThetaQuarterTurn` there. Alternatively, you can add a new `ImageViewHandler` example that demonstrates that class.

components/pango_display/src/handler_image.cpp
```diff
@@ -431,7 +543,7 @@ void ImageViewHandler::Special(View& view, pangolin::InputSpecial inType, float
         const float d[2] = {p1,p2};
         const float is[2] = {rview.x.Size(),rview.y.Size() };
         const float df[2] = {is[0]*d[0]/(float)view.v.w, is[1]*d[1]/(float)view.v.h};
-        ScrollView(-df[0], -df[1]);
```
stevenlovegrove, line 434: This looks like it changes scroll direction behaviour on MacOS? Was it behaving incorrectly for you?

nickcharron:
@stevenlovegrove is there any chance we could get this landed shortly? We are about to release some code that depends on these changes. If there is anything I can do to help speed this up, please let me know!

stevenlovegrove:
> @stevenlovegrove is there any chance we could get this landed shortly? We are about to release some code that depends on these changes. If there is anything I can do to help speed this up, please let me know!

Hi - I had a question about your ScrollView change which you can see above. That is the only thing blocking merge.

components/pango_display/src/handler_image.cpp
mp3guy: I actually encountered this recently when trying to build a GUI on top of master pangolin, it looks like you drive-by reversed this in a recent commit: https://github.com/stevenlovegrove/Pangolin/commit/69b9d866579a7e292f60be0b6eec73699896757a

With this change, panning e.g. ImageViews has vertical motion flipped. I suspect this is an OS issue and the change you made "fixed" it for Mac, but broke it for Windows/Linux.

stevenlovegrove: I see - the problem you see must be for right-click panning on win / linux? The SpecialInput method is generally only used on macos, but there is a reference to it https://github.com/stevenlovegrove/Pangolin/blob/837842ff93c0cafaa264071522a4e8ce16780524/components/pango_display/src/handler_image.cpp#L523 where it may be called on other platforms. I imagine reversing it there instead should make everything work correctly for all platforms.

SeaOtocinclus: @stevenlovegrove How do you want debug this, should we fix the behavior in the Special function you pointed us or just locally on the pointed line 434?

stevenlovegrove: Let me merge this now so you're no longer completely blocked, and then if you were able to try my suggested fix to ensure it works on Linux / Win that would be great. If not, I can try to do it later.

nickcharron: Thanks @stevenlovegrove! Much appreciated

PR #767, opened 2022-05-29
fix a few warnings.

ethanrublee:
- signed/unsigned comparison
- unsigned is always >= 0
- Intializer list order
- strncpy complains that destination is bound size of 16


stevenlovegrove:
Thanks!

PR #773, opened 2022-06-11
feat: make unattended for apt, lint of README.md, -u install

richtong:
A few fixes to make Docker installation of Pangolin work

Cleanup The current README.md now passes markdownlint

Convert install script to be unattended for the apt case so it works with Dockerfile

Updated README.md to note Ubuntu installation needs a -u first

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,15 @@
-What is Pangolin
-====================================
+# What is Pangolin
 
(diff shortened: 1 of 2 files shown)


stevenlovegrove:
Thanks for this. The script update looks great. The readme update I need to be careful with because I had trouble getting it to render correctly both on github.com and stevenlovegrove.github.io/Pangolin where it is also used. When I get a moment I might try splitting this unless you want to do so yourself.

richtong:
Happy to do :-)

Sent via Superhuman iOS ( ***@***.*** )

PR #774, opened 2022-06-20
added missing include for std::min

dcanelhas:
https://en.cppreference.com/w/cpp/algorithm/min

The image component fails to build on win 10 using cmake to generate a VS2019 solution without including algorithm.


stevenlovegrove:
Thanks!

PR #795, opened 2022-09-30
ASAN fix for range

mp3guy:
Basically float::max cast to int is too big for that type and it gets sad. Even worse, float doesn't have precision to store even close to int max, meaning you can't just do a std::min() trick to clamp the value.


stevenlovegrove:
Thanks!

PR #800, opened 2022-10-15
Fix for windows build.

stevenlovegrove:


christian-rauch:
Which issue does this fix? Is this still required?

stevenlovegrove:
We can close it. Presumably defunct after this time.

PR #831, opened 2023-01-23
Support PANGOLIN_WINDOW_URI environment variable

isherman:


stevenlovegrove (approved):
Great, thanks! I think this can be rebased against pango2

PR #833, opened 2023-02-06
pango2: layer group bug fix and  layer group / flex layout / layout constraints code comments, refactor

strasdat:
- layout bug fixes (devide by zero aspect hints => fallback to 1.0 ratio)
 - minimal test case / smoke test for ``computeLayoutConstraints``
 - minor refactor and comments for LayoutGroup and corresponding functions
   * separate LayerGroup into constants params struct + constraints struct (first layout path) + region (result of second layout pass)
   * some encapsulation of AspectRatio hint to catch NAN / devided by zero and other improper ratios systematically

### DIFF
--- a/components/pango_context/include/pangolin/gui/layer_group.h
+++ b/components/pango_context/include/pangolin/gui/layer_group.h
@@ -13,7 +13,9 @@ namespace pangolin
 ////////////////////////////////////////////////////////////////////
 /// Represents a (possibly nested) arrangement of Panels on screen
 ///
-struct LayerGroup {
+class LayerGroup
+{
+  public:
   enum class Grouping {
     stacked,     // layers blended over one another
     tabbed,      // one layer shown at a time, with user selecting current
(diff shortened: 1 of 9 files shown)


components/pango_context/include/pangolin/gui/layer_group.h
```diff
@@ -23,22 +25,52 @@ struct LayerGroup {
...
+
+  bool hasConstraint() const { return bool(constraints_); }
+
+  void setConstraints(const Constraints& constraints)
+  {
+    constraints_ = constraints;
+  }
+
+  // Precondition: this->hasConstraint() must be true
+  const Constraints& constraints() const { return FARM_UNWRAP(constraints_); }
```
stevenlovegrove, line 56: I think it's best to just return a const ref to the optional instead of having a hasConstraints() and constraints() method.

components/pango_context/include/pangolin/gui/layer_group.h
```diff
@@ -23,22 +25,52 @@ struct LayerGroup {
...
+
+  const std::vector<LayerGroup>& children() const { return children_; }
+  std::vector<LayerGroup>& mutChildren() { return children_; }
+
+  bool hasRegion() const { return bool(region_); }
+
+  void setRegion(const Region2I& region) { region_ = region; }
```
stevenlovegrove, line 63: and here

components/pango_context/include/pangolin/gui/layer_group.h
```diff
@@ -23,22 +25,52 @@ struct LayerGroup {
...
+  void setConstraints(const Constraints& constraints)
+  {
+    constraints_ = constraints;
+  }
+
+  // Precondition: this->hasConstraint() must be true
+  const Constraints& constraints() const { return FARM_UNWRAP(constraints_); }
+
+  const std::vector<LayerGroup>& children() const { return children_; }
+  std::vector<LayerGroup>& mutChildren() { return children_; }
```
stevenlovegrove, line 59: the mut prefix isn't a pattern I really use in pangolin.

stevenlovegrove (changes requested):
This is great, thanks Hauke! It can't have been pleasant to work through, I know this code was a bit rough!

Just a few very small comments. I haven't had a chance to run it but it looks good.

Additionally, you accidentally included 14.shared-context.cpp in your PR.

components/pango_context/include/pangolin/gui/layer_group.h
strasdat: good idea will done

components/pango_context/include/pangolin/gui/layer_group.h
strasdat: done

components/pango_context/include/pangolin/gui/layer_group.h
strasdat: sure, done

stevenlovegrove:
This has also been cherry-picked in already. Thanks!

PR #836, opened 2023-02-10
Pango2 to merge

stevenlovegrove:


stevenlovegrove:
@strasdat @isherman some good fixes in here and I want to highlight a small API change for DeviceBuffer in case it causes any problems. This also adds a way to get keyboard events on top of the recent double-click cursor callbacks which might be handy.

PR #841, opened 2023-02-28
feat: new composable group api in sophus2

strasdat:


stevenlovegrove:
Cool - you managed to keep the API pretty much unchanged! Looks like you just have to update the Sophus hash in the github workflow to fix the build.

PR #854, opened 2023-04-01
Fix Icosphere

isherman:
I believe the issue with the current implementation is that when indices are specified as a `std::vector<Eigen::Matrix<uint32_t, 3, 1>>`, we run into Eigen alignment issues, i.e. http://eigen.tuxfamily.org/dox-devel/group__TopicFixedSizeVectorizable.html.

I'm not sure what deeper fixes might be necessary to prevent, or fix that usage in the API, but this is a solution for the Icosphere at least; indices are specified as `std::vector<unsigned int>`.

before and after, white icosphere with 4 subdivisions next to default cube:
before
after

### DIFF
--- a/components/pango_context/src/make_drawable.cpp
+++ b/components/pango_context/src/make_drawable.cpp
@@ -143,8 +143,7 @@ Shared<Drawable> DrawableConversionTraits<draw::Icosphere>::makeDrawable(
   auto prims = DrawnPrimitives::Create({
       .element_type = DrawnPrimitives::Type::triangles,
   });
-  using FaceT = Eigen::Matrix<uint32_t, 3, 1>;
-  std::vector<FaceT> faces;
+  std::vector<uint32_t> faces;
 
   // http://www.songho.ca/opengl/gl_sphere.html#icosphere
   static float constexpr kHAngle = M_PI / 180 * 72;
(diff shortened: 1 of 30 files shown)


components/pango_context/src/make_drawable.cpp
```diff
@@ -143,87 +143,76 @@ Shared<Drawable> DrawableConversionTraits<draw::Icosphere>::makeDrawable(
...
+
+    indices.swap(new_indices);
   }
 
+  auto colors = std::vector<Color>(vertices.size(), sphere.color);
+
   prims->vertices->queueUpdate(std::move(vertices));
-  prims->indices->queueUpdate(std::move(faces));
+  prims->indices->queueUpdate(std::move(indices));
+  prims->colors->queueUpdate(std::move(colors));
```
stevenlovegrove, line 215: Rather than add a vbo of duplicate entries like this, you should be able to just set default_color (https://github.com/stevenlovegrove/Pangolin/blob/pango2/components/pango_context/include/pangolin/drawable/drawn_primitives.h#L112)

stevenlovegrove:
Great, thanks Ian!

components/pango_context/src/make_drawable.cpp
isherman: thanks, done

isherman:
Needs another fix, please hold off on merging.

PR #864, opened 2023-05-05
Add env var to skip binding ESC to quit Pangolin

ChristopherRabotin:
Setting PANGOLIN_BIND_ESC_TO_QUIT to anything but `1` will skip binding the escape key to the quit function of Pangolin. For backward compatibility, not setting this environment variable will keep the previous behavior, i.e., pressing escape will quit pangolin

### DIFF
--- a/components/pango_display/src/display.cpp
+++ b/components/pango_display/src/display.cpp
@@ -182,7 +182,13 @@ void RegisterNewContext(const std::string& name, std::shared_ptr<PangolinGl> new
     context = newcontext.get();
 
     // Default key bindings can be overridden
-    RegisterKeyPressCallback(PANGO_KEY_ESCAPE, Quit );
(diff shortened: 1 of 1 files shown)


stevenlovegrove:
Thanks for this PR, Christopher. Unfortunately, the version of Pangolin that we're using is on a different branch named pango2 which works a bit differently. On that branch we can probably just remove this default for now (cc @isherman)

ChristopherRabotin:
Hey Steve, any update on getting this done for version 2? We're still encountering the escape key issue. If there was just a way to turn it off by env var, that would be really good and straightforward.

isherman:
@ChristopherRabotin this has been addressed on the `pango2-dev` branch, going to close here per @stevenlovegrove's suggestion

PR #866, opened 2023-05-10
[Build] Fix for Fedora38 and modern G++

SeaOtocinclus:
- Fedora38 is using g++ (GCC) 13.1.1 and it was complaining about a missing definition error:
```
‘uint32_t’ does not name a type
    7 | using PangoTagType = uint32_t;
```
We are fixing it here.


stevenlovegrove:
Thanks!

SeaOtocinclus:
You're welcome!

PR #868, opened 2023-05-16
Add inline to varextra's checkReset()

isherman:
Fixes a linker error 
```
[build] /usr/bin/../lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/bits/deque.tcc:770: multiple definition of pangolin::checkReset();
```

CI was failing due to pre-commit format check; I pushed a fix for that to this PR as well.


stevenlovegrove (approved):
Thanks!

PR #870, opened 2023-05-30
Fix text renderings in 3D

### LINKED ISSUE #776: GlFont doesn't display in 3D
Hello, I got a text display problem in 3D enviroment.
I used to use
`pangolin::GlFont::I().Text(name).Draw(x, y, z)` for adding some information on 3d points. 
After upgrade Pangolin to 0.8 version, `GlFont()::I()` is not avaliable, so I use `pangolin::default_font().Text(name).Draw(x, y, z)` instead, but I cannot get any text display on 3D point any more. Can anyone help solve this?

mp3guy:
This fixes #776.


stevenlovegrove:
Thanks @mp3guy!

PR #875, opened 2023-06-15
Small fix for drag selection not being correct on mouse up event.

stevenlovegrove:


Hackerman342:
@stevenlovegrove can you merge this into `pango2` so I can use both the fixes here and those in `pango2` in a submodule?

Issue #736, opened 2022-02-16
Dependencies for Archlinux


niu-niu:
Is it possible to add dependencies list for archlinux platform?

stevenlovegrove:
I don't have any experience with archlinux, but If you wanted to create a pull request with the required changes in install_prerequisites.sh I'd consider including them

PeterBee97:
pangolin-git is in AUR already. However the build also fails with ffmpeg 5.0, as mentioned in #737 .

PeterBee97:
I patched the CMakeList.txt to find the ffmpeg4.4 files and it works.

christian-rauch:
@niu-niu You can extend the [install script](https://github.com/stevenlovegrove/Pangolin/blob/master/scripts/install_prerequisites.sh) with an additional package manager. For this, you have to add `pacman` to `MANAGERS=(dnf apt port vcpkg brew)` and then add a new else block with the package list to list of package manager cases:
https://github.com/stevenlovegrove/Pangolin/blob/d64507241dc79ecd98c1dea267d74dac34663629/scripts/install_prerequisites.sh#L114-L123.

PR #896, opened 2023-12-20
Modify install_prerequisites.sh to support pacman (Arch)

ertsth:
Closes https://github.com/stevenlovegrove/Pangolin/issues/736 (this is an old one, but I think having pacman support would be great anyway).


stevenlovegrove:
Thanks!

PR #930, opened 2024-05-22
Add SONAME Versioning

damaestro:
- Add SOVERSION to all SOs

This has only been tested on Linux.

### DIFF
--- a/components/pango_python/CMakeLists.txt
+++ b/components/pango_python/CMakeLists.txt
@@ -54,6 +54,11 @@ if(BUILD_PANGOLIN_PYTHON AND Python_FOUND AND pybind11_FOUND)
     target_link_libraries(${COMPONENT} PUBLIC
         pango_core pango_display pango_plot pango_video pango_vars
     )
+
+    set_target_properties(
+        ${COMPONENT} PROPERTIES VERSION ${PANGOLIN_VERSION} SOVERSION ${PANGOLIN_VERSION_MAJOR}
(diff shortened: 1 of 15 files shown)


components/pango_python/CMakeLists.txt
```diff
@@ -73,6 +78,12 @@ if(BUILD_PANGOLIN_PYTHON AND Python_FOUND AND pybind11_FOUND)
         $<INSTALL_INTERFACE:include>
     )
 
+    if (NOT APPLE)
+        set_target_properties(
+            pypangolin PROPERTIES VERSION ${PANGOLIN_VERSION} SOVERSION ${PANGOLIN_VERSION_MAJOR}
+        )
+    endif()
+
```
christian-rauch, line 86: Does this actually have an effect? Looking at the log, the generated so file:
> Linking CXX shared module pypangolin-0.9.1.data/purelib/pypangolin.cpython-310-x86_64-linux-gnu.so

does not contain the versioning. Additionally, this should not be required for the `pypangolin.cpython` since this is not dynamically linked. Are the `PROPERTIES VERSION` really required in the `pypangolin` component?

damaestro: @christian-rauch ah, you might be right. The pypangolin library does not need to be, but the embedded python does. I'll push an update after doing a test build.

christian-rauch (approved):

Issue #939, opened 2024-06-14
Font rendering issue after update


mateosss:
When updating from 86eb4975 to one of the latest master commits 9d8fef79 in my project fonts look off. What could be the reason?

![before](https://github.com/stevenlovegrove/Pangolin/assets/5622899/cbe4c70a-a84c-4ead-b76d-949f335a4595)

![after](https://github.com/stevenlovegrove/Pangolin/assets/5622899/c4326cf6-bc80-4f03-b0a2-09b8697a50bb)

christian-rauch:
There are 4 years of updates between these two commits. You would need to do a git bisect to find the commit that introduced the font rendering issue.

christian-rauch:
After a git bisect I found that the last good commit is 3a1d89eb and that the next commit cba1ef20 introduces the issue.

| good (3a1d89eb)  | bad (cba1ef20) |
| ------------- | ------------- |
| ![pangolin_good_3a1d89eb](https://github.com/stevenlovegrove/Pangolin/assets/8226248/279e8b9c-a67c-4d28-a207-87a4ddff7023)  | ![pangolin_bad_cba1ef20](https://github.com/stevenlovegrove/Pangolin/assets/8226248/fe32c8a7-ed00-462c-b31c-700ccb8bd7c3) |

PR #940, opened 2024-06-16
Revert "pango_display: less state changes in widgets"

christian-rauch:
This reverts commit cba1ef20254718ff08cb2b84bc825a2b531d957a. Fixes https://github.com/stevenlovegrove/Pangolin/issues/939.

| current master  | PR |
| ------------- | ------------- |
| pango_sd_master | pango_sd_pr |
| pango_hello_master | pango_hello_pr |


christian-rauch:
@stevenlovegrove The "continuous coords" from the commit seem to blur the rendered text (https://github.com/stevenlovegrove/Pangolin/issues/939). What was the reason for changing this? Can this be safely reverted or are we going to introduce issues somewhere else?

mateosss:
Thanks for the update! I can confirm this fixes #939

stevenlovegrove:
@christian-rauch looking at the commit, it seems I made an error and should have used `ProjectionMatrixOrthographic(-0.5, w-0.5, -0.5, h-0.5, -1.0, 1.0).Load();` instead of the continuous coords.

The reason for the commit was to avoid the extra glMatrixMode calls (since we have to store the projection and model view matrices and then set them in a different method https://github.com/stevenlovegrove/Pangolin/blob/13e4a1804cefe8ddfd22cdddfac183d5d2828859/components/pango_opengl/src/viewport.cpp#L64).

Reverting is safe. We could also replace the ProjectionMatrixOrthographic call with the correct pixel centric version to avoid the two extra GL API calls.

christian-rauch:
> @christian-rauch looking at the commit, it seems I made an error and should have used `ProjectionMatrixOrthographic(-0.5, w-0.5, -0.5, h-0.5, -1.0, 1.0).Load();` instead of the continuous coords.

Thanks. This seems to work. I updated the PR.

@mateosss Can you check again with the new commit in the PR?

PR #953, opened 2024-09-24
Improve pacman prerequisites installation

tmmsartor:
Avoid reinstalling already up to date packages


christian-rauch:
For completeness, the [ArchWiki](https://wiki.archlinux.org/title/Pacman) says for `pacman -S gnome`:
> Note: If a package in the list is already installed on the system, it will be reinstalled even if it is already up-to-date. This behavior can be overridden with the `--needed` option.

Issue #985, opened 2025-06-04
Suggestion for M1/macOS Clang compatibility


ZHLRJ:
### Fix: Clang error on macOS due to VLA usage

When building Pangolin on macOS with Apple Clang (especially on M1/ARM Macs), compilation fails due to the use of variable length arrays (VLAs), which are not standard C++ and are treated as a Clang extension:

![Image](https://github.com/user-attachments/assets/7ef21d91-a36a-42e2-8afc-dfa3f126d8f9)

![Image](https://github.com/user-attachments/assets/ba194a17-bbe8-4f3b-b136-085e6eb949ff)

error: variable length arrays in C++ are a Clang extension [-Wvla-cxx-extension]

A temporary workaround is to suppress the warning using:

```cmake
add_compile_options(-Wno-vla-extension)

if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
    add_compile_options(-Wno-vla-extension)
endif()

christian-rauch:
> When building Pangolin on macOS with Apple Clang (especially on M1/ARM Macs), compilation fails due to the use of variable length arrays (VLAs), which are not standard C++ and are treated as a Clang extension:

The CI tests macOS 14 with `AppleClang 15.0.0.15000309` on the [`macos-14` `arm64` runner](https://github.com/actions/runner-images/blob/main/images/macos/macos-14-Readme.md) and does not trigger the error. Can you reproduce this on the CI? Which macOS and compiler version are you using?

> A temporary workaround is to suppress the warning using:

Since all the values are known at compile time, I am wondering if using `constexpr` would also circumvent the issue.

ZHLRJ:
I'm on macOS 15.5 with Apple clang-1700.0.13.5, which seems stricter about VLAs. Your CI uses Clang 15, so that explains the difference.

PR #986, opened 2025-06-09
Fix for variable length array initialization error.

YLouWashU:
This PR addresses the following issue:

On Mac-ARM64 computer when building from source, I got the following errors: 
```
/tmp/Pangolin/components/pango_display/src/view.cpp:315:16: error: variable length arrays in C++ are a Clang extension [-Werror,-Wvla-cxx-extension]
  315 |     GLfloat zs[zsize];
      |                ^~~~~
/tmp/Pangolin/components/pango_display/src/view.cpp:315:16: note: initializer of 'zsize' is not a constant expression
/tmp/Pangolin/components/pango_display/src/view.cpp:314:15: note: declared here
  314 |     const int zsize = zl*zl;
      |               ^
1 error generated.
```
This is caused by compiler not allowing creation of variable length array. 

This PR addresses this by creating a pointer that points to the data member of a std::vector.

### DIFF
--- a/components/pango_display/src/view.cpp
+++ b/components/pango_display/src/view.cpp
@@ -312,7 +312,8 @@ GLfloat View::GetClosestDepth([[maybe_unused]] int x, [[maybe_unused]] int y, in
 #endif
 
     const int zsize = zl*zl;
-    GLfloat zs[zsize];
+    std::vector<GLfloat> zsVec(zsize);
+    GLfloat* zs = zsVec.data();
 
 #ifndef HAVE_GLES
     glReadBuffer(GL_FRONT);


christian-rauch:
Would this also work if you use `constexpr`? And could you update the macOS runner first to reproduce the issue you are trying to fix?

christian-rauch:
I assume you are trying to fix https://github.com/stevenlovegrove/Pangolin/issues/985. There, we hypothesised that this is likely caused by newer macOS / AppleClang version. The CI is using macOS 14 and the issue seems to appear on macOS 15. So updating the CI pipeline here:
https://github.com/stevenlovegrove/Pangolin/blob/122bb3ee8b8b06761e9738e6bb747f81bfa74002/.github/workflows/build.yml#L19
should reproduce the issue and your PR should fix the issue.

components/pango_display/src/view.cpp
```diff
@@ -312,7 +312,8 @@ GLfloat View::GetClosestDepth([[maybe_unused]] int x, [[maybe_unused]] int y, in
 #endif
 
     const int zsize = zl*zl;
-    GLfloat zs[zsize];
+    std::vector<GLfloat> zsVec(zsize);
+    GLfloat* zs = zsVec.data();
```
christian-rauch, line 316: Why do you allocate the memory through a `std::vector` and not `malloc`/`free`? Also, when you turn this into a dynamic array, you can remove the `MSVC` workaround above.

christian-rauch (changes requested):

christian-rauch:
I updated the runner, enabled VLA errors to have this triggered when new VLAs are added, and I replaced the `zs` VLA as suggested with a `std::vector<>` in https://github.com/stevenlovegrove/Pangolin/pull/987. Can you check that PR if it solves your issue?

PR #992, opened 2025-08-24
Multiple textures

tzukpolinsky:
Hi
this is the code that enables multiple textures
As you can see, it mainly takes the current logic and put it into a for loop.

i hope that you would accept the shader.h as well. when i started working with shaders and didnt know anything about it, this examples really helped me and i hope they would help others in the future :)

PR #860, opened 2023-04-16
working with multiple textures, and some more readable examples for shaders

tzukpolinsky:
Hello
For an academic project I'm creating in Haifa University, I tried to use the ModelViewer tool to load multiple textures.
In the first attempt, because in geometry_obj the texture names are given using an iterator and in the shader there is only one texture "texture_0", I couldn't use the program as is.

So I attached the texture names to the object names,and in GLDraw changed the structure of the function.

I hope you could integrate it in the repository, because loading and working with several textures is a painful subject across the web.

Thank you in advance
Tzuk Polinsky

PR #995, opened 2025-08-25
working with multiple textures, rebasing current pangolin version

christian-rauch:
This PR is a cleaned up version of https://github.com/stevenlovegrove/Pangolin/pull/860 and https://github.com/stevenlovegrove/Pangolin/pull/992.


tzukpolinsky:
I need to install the system on my computer, it would take me several days, but i would run the examples and see they all work. And try to find a model somewhere with multiple textures.

